PR #12, opened 2021-06-18
Upgrade the HMAC Relay to protect Raptoreum SmartNodes Smart Contracts

Gapcoin1:
Upgrading the HMAC keys to 512 Bits from 128 Bits will provide security to protect the non-tampered flow of Smart Contracts and Assets through SmartNodes

Making Man-In-The-Middle attacks to tamper much more difficult!

### DIFF
--- a/src/crypto/hmac_sha512.cpp
+++ b/src/crypto/hmac_sha512.cpp
@@ -8,22 +8,22 @@
 
 CHMAC_SHA512::CHMAC_SHA512(const unsigned char* key, size_t keylen)
 {
-    unsigned char rkey[128];
-    if (keylen <= 128) {
+    unsigned char rkey[512];
+    if (keylen <= 512) {
         memcpy(rkey, key, keylen);
-        memset(rkey + keylen, 0, 128 - keylen);
+        memset(rkey + keylen, 0, 512 - keylen);
     } else {
         CSHA512().Write(key, keylen).Finalize(rkey);
         memset(rkey + 64, 0, 64);
     }
 
-    for (int n = 0; n < 128; n++)
+    for (int n = 0; n < 512; n++)
         rkey[n] ^= 0x5c;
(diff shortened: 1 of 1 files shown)


Bigpiggy01:
Hi there sorry about the slow response been going at it hard to get a new version ready and this got sidetracked somewhat.
Why upgrade this across the board and not just as a helper for rpc calls?
Also, could you please elaborate slightly on the form of such a mitm attack? As I see it most of the concerns would center around upload if that is designated by an additional quorum, I do not see how a traditional mitm attack could be set up for it.

Gapcoin1:
I don't know how much I can simplify it.  But yes, essentially using 512 bit keys would ensure that any uploads would be to the correct server......and that each Node was working with the correct Chain!

It's like the Russian capsule dolls.  The HMAC is the outer layer that relays the Current state, but mostly it does not become part of the Blockchain.....it encapsulates the SHA256 Blockchain, and Private Keys.  

Of course it's not so much an encapsulation process, but more of a Train Cars tied together.  The HMAC is the hardest part of the Train Set.........SHA-512, just as Satoshi original design.  He only used OpenSSL and back in the day everything was only 32 Bits.......

Sathoshi always designed the Sha-512 HMAC keys to be the leverage point to easily upgrade the protection for the entire Blockchain.........From an Encryption breaking point of view, but Bitcoin was broken early on from a different Flaw where actors were able to generate a couple hundred thousand or more Bitcoins which required a hard-fork to correct!

HMAC is more critical in an environment using SmartNodes for obvious reasons!  HMAC is the Front Engine of the Chain, but leaves not much trace of it's work, other than getting it's Cargo to it's destination safely, and the Cargo is the only thing that gets measured!

Bigpiggy01:
We will do some testing post release as I am not sure exactly how much of an impact upgrading the wallets etc will have on the overall eco system.

PR #14, opened 2021-07-12
Upgrade the HMAC Relay to protect Raptoreum SmartNodes Smart Contracts

Gapcoin1:
Upgrading the HMAC keys to 512 Bits from 128 Bits will provide security to protect the non-tampered flow of Smart Contracts and Assets through SmartNodes

Making Man-In-The-Middle attacks to tamper much more difficult!

However, I am not sure since SipaHash if the Bitcoin code even calls this function but only the HMACSha512.h

### DIFF
--- a/src/crypto/hmac_sha512.cpp
+++ b/src/crypto/hmac_sha512.cpp
@@ -8,22 +8,22 @@
 
 CHMAC_SHA512::CHMAC_SHA512(const unsigned char* key, size_t keylen)
 {
-    unsigned char rkey[128];
-    if (keylen <= 128) {
+    unsigned char rkey[512];
+    if (keylen <= 512) {
         memcpy(rkey, key, keylen);
-        memset(rkey + keylen, 0, 128 - keylen);
+        memset(rkey + keylen, 0, 512 - keylen);
     } else {
         CSHA512().Write(key, keylen).Finalize(rkey);
         memset(rkey + 64, 0, 64);
     }
 
-    for (int n = 0; n < 128; n++)
+    for (int n = 0; n < 512; n++)
         rkey[n] ^= 0x5c;
-    outer.Write(rkey, 128);
+    outer.Write(rkey, 512);
 
-    for (int n = 0; n < 128; n++)
+    for (int n = 0; n < 512; n++)
         rkey[n] ^= 0x5c ^ 0x36;
-    inner.Write(rkey, 128);
+    inner.Write(rkey, 512);
 }
 
 void CHMAC_SHA512::Finalize(unsigned char hash[OUTPUT_SIZE])


salmanmarvasti:
how does this work  with the Future contracts under test... dev commits has dried up with the rise in the RTM price?

Bigpiggy01:
Futures are the the protocol level not contract level.

As to commits check the right branch and requests.

JSanchezFDZ (changes requested):
Reviewed `adf65dde7624eb71c613b415ba31d27713aa469a`: the complete HMAC change, SHA-512 implementation, HD-key callers, and existing test vectors; also compared the checked `develop` snapshot. **Requesting changes: retain the original HMAC construction.**

**P1 — Increasing the block from 128 to 512 bytes breaks HMAC-SHA512 and HD compatibility.**

At [hmac_sha512.cpp:11–26](https://github.com/Raptor3um/raptoreum/blob/adf65dde7624eb71c613b415ba31d27713aa469a/src/crypto/hmac_sha512.cpp#L11), `128` is SHA-512's block size **in bytes**, not a 128-bit key length. The existing class already produces a 64-byte/512-bit digest. HMAC pads must use the underlying hash's block size; expanding them and changing the long-key threshold produces a different algorithm. See [RFC 2104 §2](https://www.rfc-editor.org/rfc/rfc2104.html#section-2) and the [RFC 4231 vectors](https://www.rfc-editor.org/rfc/rfc4231.html).

**Direct verification:** I compiled the actual unmodified `hmac_sha512.cpp`/`sha512.cpp` from the merge base, this HEAD, and `develop` with GCC 13.3/C++17. All six RFC vectors already present in the repository pass on base/develop; all six fail on HEAD. For the public `Hi There` vector, the correct prefix is `87aa7cdea5ef619d`; HEAD returns `4213fe46f6f1b389`. Python's standard `hmac`/`hashlib.sha512` independently agrees with the expected vectors.

| Implementation | Six RFC vectors | Master HMAC with public BIP32 test seed |
| --- | --- | --- |
| Merge base | 6/6 correct | Correct |
| PR HEAD | 0/6 correct | Different |
| Checked develop | 6/6 correct | Correct |

This is used by `CExtKey::SetMaster` and `BIP32Hash`, including private/public child derivation. `CHDChain::DeriveChildExtKey` is also called when `CWallet::GetKey` retrieves private keys for existing addresses. The same seed/path therefore derives different keys, breaking compatibility with existing wallets and BIP32. I traced the signing/key-retrieval consequences statically; I did not use a real wallet or perform transaction signing. The additional public-seed check exercises the master-generation HMAC, not the complete BIP32 suite.

The `keylen > 512` branch also leaves `rkey[128..511]` uninitialized after writing the digest and only 64 zero bytes. This is a static finding with no identified production caller supplying such a long key. Fixing only that initialization would still leave the algorithm incompatible.

**Correction and validation before integration**

Keep the original 128-byte block, threshold, and pad lengths. Run the existing `crypto_tests/hmac_sha512_testvectors` and `bip32_tests`; do not replace their expected values to accommodate this change. Correct the description's distinction between block size, output size, and caller-provided key length. No demonstrated protocol issue or performance benefit justifies changing this standardized primitive.

The isolated compilations/executions used a container without network access and public vectors only. No full node build, complete BIP32 suite, wallet functional test, or benchmark was executed in this review.

PR #21, opened 2021-08-20
Ft/future

socialruins:


src/qt/transactionrecord.cpp
```diff
@@ -26,6 +28,85 @@ bool TransactionRecord::showTransaction(const CWalletTx &wtx)
...
+/* Return block height for transaction */
+int TransactionRecord::getTransactionBlockHeight(const CWalletTx &wtx)
+{
+    // Find the block the tx is in
+    CBlockIndex* pindex = nullptr;
+    BlockMap::iterator mi = mapBlockIndex.find(wtx.hashBlock);
+    if (mi != mapBlockIndex.end())
+        pindex = (*mi).second;
+
+    int txBlock = pindex ? pindex->nHeight : std::numeric_limits<int>::max();
```
npq7721: if pidex is null then transaction is not yet confirm so therefore it should be 0 instead of int max

npq7721: actually 0 is not entirely correct, it depend on how u use this number, the idea is pindex is null it is unconfirmed so as long u handle unconfirmed (not yet been mined), then should be good

npq7721 (approved):
thanks for the contribution

PR #23, opened 2021-09-22
CryptoNight Variants used to hash GhostRider simplified and decreased…

malbit:
… imported files. Less memory usage as well as lighter binary file.


npq7721:
look good. thanks

PR #36, opened 2021-11-05
upstream merge changes from dash v0.16.x

malbit:

PR #58, opened 2021-11-20
Protx list bug collateral update

CodeIsTheKey:
The existing logic was not handling the case when using `protx list valid {height}` properly with collateral upgrades.
1. It would see the spent output from before the upgrade and consider it invalid, when it was unspent at the provided height.
1. Even if it was unspent, it would not be considered valid if it didn't match the collateral amount at the tip.

As a test, `protx list valid true 100000`  only returns 2 records because there were two nodes with 1.25M RTM and they are still unspent.
With the updates in this PR, it returns 292 nodes.


src/validation.cpp
```diff
@@ -353,6 +353,40 @@ bool CheckSequenceLocks(const CTransaction &tx, int flags, LockPoints* lp, bool
...
+
+        if (GetSpentIndex(key, value)) {
+            if (value.blockHeight <= height) {
+                // Not Valid, already spent at this height
+                return false;
+            }
+        }
+        else
+        {
+            // Not spent, it was valid at this height
```
CodeIsTheKey: My only concern with this call is that in this case, I am returning a default Coin (since it was later spent).  I doesn't affect the caller, but I don't know if this would be a problem/confusion later.

src/evo/deterministicmns.cpp
```diff
@@ -118,15 +118,35 @@ bool CDeterministicMNList::IsMNPoSeBanned(const uint256& proTxHash) const
...
+{
+    uint256 mnHash = dmn.get()->collateralOutpoint.hash;
+    Coin coin;
+    // Should this be called directly or use pcoinsTip->GetCoin(outpoint, coin) without locking cs_main
+    bool isValidUtxo = GetUTXOCoin(dmn->collateralOutpoint, coin, height);
+    SmartnodeCollaterals collaterals = Params().GetConsensus().nCollaterals;
+
+    int64_t amount = coin.out.nValue;
+
+    if (height != chainActive.Tip()->nHeight) {
```
npq7721: it would be possible to get segfault if Tip is null which it is at block 0.

src/validation.cpp
npq7721: i would remove this else case since it have no code but comment and i would combine if GetSpendIndex and value.blockHeight <= height to be in one if with &&

src/wallet/wallet.cpp
```diff
@@ -4854,6 +4854,21 @@ void CWallet::ListLockedCoins(std::vector<COutPoint>& vOutpts) const
     }
 }
 
+void CWallet::ListProTxCoins(int height, std::vector<COutPoint>& vOutpts)
+{
+    auto mnList = deterministicMNManager->GetListForBlock(chainActive[height]);
+
+    AssertLockHeld(cs_wallet);
```
npq7721, line 4859: i would make 4861-4869 to a method so i can be reused for void CWallet::ListProTxCoins(std::vector<COutPoint>& vOutpts), they have the same code

npq7721 (commented):
please check comments and correct it

CodeIsTheKey:
I forgot to mention, this also fixes the same issue with your own SmartNodes:
`protx list wallet true 100000` will now show you SmartNodes that were valid at the given height.

Bigpiggy01:
Good stuff man, could you ping me again on Discord so I can shoot you an invite to out next developer meeting.

src/validation.cpp
```diff
@@ -353,6 +353,35 @@ bool CheckSequenceLocks(const CTransaction &tx, int flags, LockPoints* lp, bool
     return EvaluateSequenceLocks(index, lockPair);
 }
 
+bool GetUTXOCoin(const COutPoint& outpoint, Coin& coin, int height)
+{
+    LOCK(cs_main);
+    bool gotCoin = pcoinsTip->GetCoin(outpoint, coin);
+
+    if (height != chainActive.Tip()->nHeight) {
```
npq7721: still potential segfault here if chainActive.Tip() is nullptr

malbit:
It might resolve issue with testnet base sync cus above block_height 160k I constantly getting connection_refused (111) from nodes at #36 
Will do merging and check then

CodeIsTheKey:
I am going to move this over to the develop branch and re-issue the PR.

PR #61, opened 2021-11-21
Libraries, update boost.mk endpoint from retired bintray to jfrog.

yoka:
More info here: https://www.boost.org/users/news/boost_has_moved_downloads_to_jfr.html


npq7721 (approved):
Thanks

PR #67, opened 2021-11-26
Bug/protx list with collateral update

### LINKED ISSUE #34: protx list does not handle collateral changes
### Describe the issue
When requesting the list of valid protx at a specific height, it assumes the current block height to determine required collateral

### Can you reliably reproduce the issue?
Yes

#### If so, please list the steps to reproduce below:

From the cli, issue the command: `protx list valid false 100000`

### Expected behaviour
I would expect a list of all protx's that had 600K collateral (or a higher level) at that time.

### Actual behaviour
I receive a list of 4 protxs and they all had collateral of 1,000,000 or 1,250,000.

### Screenshots.
image

### What version of Raptoreum Core are you using?
Raptoreum Core:1.2.15.2

CodeIsTheKey:
Re-made PR (https://github.com/Raptor3um/raptoreum/pull/58) to merge into develop.  No code changes since that PR.
Fixes #34.

### DIFF
--- a/src/evo/deterministicmns.cpp
+++ b/src/evo/deterministicmns.cpp
@@ -128,7 +128,8 @@ bool CDeterministicMNList::IsMNValid(const CDeterministicMNCPtr& dmn, int height
 
     int64_t amount = coin.out.nValue;
 
-    if (height != chainActive.Tip()->nHeight) {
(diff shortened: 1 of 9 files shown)


CodeIsTheKey:
BTW, this is difficult to test on the testnet - as far as I can tell, we didn't have any SmartNodes before the final collateral upgrade to 60K.

npq7721 (approved):
LGTM

Issue #38, opened 2021-11-08
Future Amount incorrect in transaction data in QT (testnet)


Raptor3um:
### Reproduce

- Receive a future transaction
- Navigate to Transactions
- Double click transaction
- Check "Future Amount"
- 
![future_transaction_amount_mismatch](https://user-images.githubusercontent.com/47279850/140764946-0dc09927-54e0-41bc-97d2-e674f23557fd.png)

CodeIsTheKey:
This will be fixed with PR #73

npq7721:
fixed in develop

PR #73, opened 2021-12-05
Bug/watch only transactions

### LINKED ISSUE #29: Not showing transactions from watch-only wallet to spendable wallet.
### Describe the issue
There seems to be a problem with how the Transaction list is displayed.
Currently, I have my own wallet and recently added an address from another wallet as Watch-Only. I get all transactions in and out from the watch-only address with one exception. When there is a transaction FROM the watch-only wallet TO any wallet that I have access to in my wallet (like the one I use to mine with) it is skipped and is not displayed. It shows neither as a transaction FROM watch only to address X (so -xyz ERM balance), nor as transaction that address X received anything from watch-only address (+xyz RTM to the wallet).
Those transactions 100% exist and are properly there when I check the explorer. Here is one of them from explorer https://explorer.raptoreum.com/tx/d6e97651e8133e56530ed15b53a53dec81db34431e9574da56170f48f1538c11 and a screenshot below from what I can see in the wallet. Screenshot from wallet is missing `RQKcAZBtsSacMUiGNnbk3h3KJAN94tstvt` and `RS48umkD6JmV2GPYmQSX5dUZShNcQW4HCP`
I have to say it looks very similar to how it looks when I send RTM from all my wallets to myself when I compress payouts and only see mini transaction fee I paid in the transaction list, not -X from a wallet or +X to another wallet.

#### If so, please list the steps to reproduce below:
1. Have one wallet X with your address
2. Have wallet Y with another address
3. Add address from wallet Y to wallet X as Watch-only
4. Send some RTM from Y (from address added in step 3) to address on wallet X
5. It should not show up in Wallet X

### Expected behavior
I think both transactions should be present in the Transaction tab. One that goes out of the Watch-Only wallet to wallet X in the wallet as well as that Wallet X received money (or at least this one if both are not possible).

### Actual behavior
The transaction is not present at all.

### Screenshots.
image

### What version of Raptoreum Core are you using?
ubuntu 18 packages from official releases, version 1.2.15.2

CodeIsTheKey:
Reworked transaction records - improves watched address handling and future transactions.  Fixes #29.

Watched addresses act like a secondary, non-spendable wallet.  So, sending a transaction from your wallet to a watched address will generate two records in the transaction log - the send and the watched receive.

### DIFF
--- a/src/coins.h
+++ b/src/coins.h
@@ -38,7 +38,8 @@ public:
     //! at which height this containing transaction was included in the active block chain
     uint32_t nHeight : 31;
 
-    int16_t nType;
+    // At this fault this is a normal transaction which is nType = 0
+    int16_t nType = 0;
 
     std::vector<uint8_t> vExtraPayload;
 
--- a/src/future/utils.h
+++ b/src/future/utils.h
@@ -12,7 +12,7 @@ class Coin;
 class COutPoint;
 class CBlockIndex;
 
-void maybeSetPayload(Coin& coin, const COutPoint& outpoint, const int16_t& nType, const std::vector<uint8_t>& vExtraPayload);
+void maybeSetPayload(Coin& coin, const COutPoint& outputIndex, const int16_t& nType, const std::vector<uint8_t>& vExtraPayload);
 //const char *validateFutureCoin(const std::vector<uint8_t>& payload, int maturity, uint32_t confirmedTime);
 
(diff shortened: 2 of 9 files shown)


CodeIsTheKey:
Here are a few examples:
image

CodeIsTheKey:
I just tested and this cleanup of the transaction records also fixes #38 
image

npq7721:
this is included in PR #77

PR #93, opened 2021-12-29
Fixed transaction time when scanning old blocks.  Fixes #91

### LINKED ISSUE #91: Watched Addresses Have Incorrect Timestamp in Qt Wallet
Using the core wallet, if you add a watched address (importaddress) - when it rescans for existing transactions, the existing transactions all get the same timestamp (when the rescan was done).  Later transactions received the expected timestamps.

Raptoreum Core version 1.2.15.3.

CodeIsTheKey:
This fixes #91.

I eventually tracked down the source of the problem.  Once found, I searched the Bitcoin repo and found the PR that fixes the same issue - https://github.com/bitcoin/bitcoin/pull/20591.

The original logic preserved the transaction order - which is good for incoming transactions, but does not work when importing pre-existing transactions.

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -1096,7 +1096,7 @@ void CWallet::MarkDirty()
     fAnonymizableTallyCachedNonDenom = false;
 }
 
-bool CWallet::AddToWallet(const CWalletTx& wtxIn, bool fFlushOnClose)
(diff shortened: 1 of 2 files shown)


CodeIsTheKey:
OBE - this was done on the pre-merge develop branch.  It has been reworked in a separate PR.

PR #95, opened 2021-12-30
Added removeaddress console command

### LINKED ISSUE #94: Add Ability to Remove Watch-Only Addresses
We have the ability to add a watch-only address using importaddress.  It would be very useful to be able to remove that address as well.  Sometimes we add the wrong address, a temporary address, or one that is no longer used/needed.

CodeIsTheKey:
This allows the user to remove a watch-only address, closes #94


CodeIsTheKey:
Rworking for new development.

PR #100, opened 2022-01-07
Update expat.mk

0x00trust:


JSanchezFDZ (commented):
Reviewed `b7cffe89dc68e810d16f5347d9db0c89e8d9b266`: both files in the original diff, download/extraction/configure helpers and dependency consumers, compared with `develop` at `d02e91323ace4ed0815a1e7c6aa168b5e30789cf`. **Recommend closing as superseded, or identifying the remaining intended change before rebasing.**

`develop` already contains Expat 2.4.1, the same archive and SHA-256, plus subsequent static-build/configuration changes. The issue template is byte-for-byte identical to the current target, and its commit is already in shared history. I found no confirmed new correctness defect in the historical version/download/hash update; these changes no longer provide a distinct update against the checked target.

**Validation performed**

- Independently downloaded the [official Expat 2.4.1 archive](https://github.com/libexpat/libexpat/releases/tag/R_2_4_1): 445,024 bytes, SHA-256 `cf032d0dba9b928636548e32b327a2d66b1aab63c4f4a13dd132c2d1d2f2fb6a`, matching the recipe. The literal helper-produced URL was also checked.
- Inspected archive layout, generated configure/Makefile files, pkg-config template, public header, and recognition of the retained static-library option.
- Artifact/layout and target-branch comparison assertions passed. Avoid replacing the current recipe with the old one and losing its later changes when resolving a rebase.

No complete depends build, downstream Qt link, cross-platform build, or runtime test was performed. These checks establish integrity and obsolescence of this proposed update, **not** the security suitability of Expat 2.4.1 today.

Optional separate documentation improvement: give security reporters a specific private endpoint or linked security policy instead of requiring them to identify the Core/Developer Teams. That wording already exists in the target and is not a new blocker here.

PR #101, opened 2022-01-07
Update qt.mk

0x00trust:


Bigpiggy01:
We're in the process of moving to Dash 17 so this is not as relevant as when it was made! Apologies for the slow review!

PR #111, opened 2022-02-05
Create application.UI.settings.html

zachwylde00:


Bigpiggy01 (changes requested):
IDE files do not particularly belong on github and for a project like ours neither do any of the third party services this relies on.

zachwylde00:
My bad. I apologize. I simply wanted to.make a mobile wallet so I wouldn't
have to be at home access this particular crypto. If you already have
something in place that would allow me access to my wallet via mobile
device, I would be thrilled to know. Thank you for taking the time to
respond and inform me of your procedures and policies. If I should do
anything else I will contact you prior to doing so and make sure I am
within all defined guidelines. Thank-you again.

PR #114, opened 2022-02-09
Add POW Cache support

CodeIsTheKey:
This PR expands on the POW LRU cache, saving and loading the cache from disk.  This dramatically speeds up (20-30x) re-syncing/rescanning processes and is also helpful for develop/debug testing cycles.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -68,6 +68,8 @@
 #include <llmq/quorums_signing.h>
 #include <llmq/quorums_utils.h>
 
+#include <primitives/powcache.h>
+
 #include <statsd_client.h>
 
 #include <stdint.h>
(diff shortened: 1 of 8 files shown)


src/init.cpp
```diff
@@ -2217,6 +2221,19 @@ bool AppInitMain()
         }
     }
 
+    strDBName = "powcache.dat";
+    uiInterface.InitMessage(_("Loading POW cache..."));
+    CFlatDB<CPowCache> flatdb7(strDBName, "powCache");
+    if (fLoadCacheFiles) {
```
npq7721: I think we can just load powcache file regradless of fLoadCacheFiles is true or not as this cache is not effected by chain state like other cache.

npq7721 (approved):
LGTM

src/init.cpp
npq7721: fixed in latest commited

PR #116, opened 2022-02-12
Fixed verification bug with future tx validation

CodeIsTheKey:
This was a subtle bug, but the extern was not by reference, so the activeChain was wrong.  The wrong reference was only used by the future TX verification logic.


npq7721 (approved):
LGTM

PR #120, opened 2022-02-22
libboost 1.73+ placeholders refactored due to deprecation

malbit:


CodeIsTheKey (approved):
LGTM - good to get out of the global namespace and nice way to cleanup.

malbit:
> LGTM - good to get out of the global namespace and nice way to cleanup.

Thx. I got used to simplify as much as it possible and to do not duplicate the code when can specify nested short_name which save a lot at the output binary :)

npq7721 (approved):
LGTM

PR #122, opened 2022-02-26
Minor cleanup of uint512, unit/regression test updates.

CodeIsTheKey:
Updated uint512 and SetHex operations to be more consistent with the uint256 implementations.

I made some updates to the regression and unit test setup for testing the arithmetic operations.  Further work is needed to adapt the full test suite.

I ran a full sync on the mainnet with this code with no issues.


npq7721 (approved):
LGTM

PR #127, opened 2022-03-12
Cryptonight Updates

### LINKED ISSUE #126: CN hash algorithms not working in debug builds
When building with --enable-debug, the CN algorithms are not computing the same has as the normal release build.  The core hashing algorithms are working fine, but the hash computation goes awry on the first CN hashing.

While doing some debugging, I noticed for CNDark and CNTurtlelite did not behave as expected.  I believe the problem covers all of the CN algorithms.

CodeIsTheKey:
Fixes #126 

### Fixed buffer overrun bug in CN slow hash

The variable 'b' was half the size it needed to be.  One step calls for copying the first half into the second half, which would overwrite the variable following it on the stack.  For release builds, g++ optimizes the stack allocation differently and what follows in the stack was safe to overwrite.  For debug builds, it was overwriting part of the working data for the hashing, causing the hash calculation to fail.

### Added CN tests and benchmarks

During testing, I added test for the 6 CN algorithms and used them for debugging.
These can be run like this:
`bench_raptoreum --filter='HASH_CN.*'`
`test_raptoreum --run_test=cn_tests`

### Fixed benign vector overrun

While running valgrind, I noticed that we were reading the 16th element of a 15 element vector.  The value wasn't used and vector will allocate space for the 16th element, but valgrind really didn't like it.  Cleaned up the logic to avoid the unneeded read.

### Fixed bug in powcache updates

No related to the other issue, but I wanted to get this in.  We cannot do an insert into the collection if the object already exists - the insert is ignored.  Added an erase to allow the "update" to proceed.


src/cryptonote/slow-hash.c
```diff
@@ -196,7 +196,7 @@ void cn_slow_hash(const char* input, char* output, int len, int variant, uint32_
   union cn_slow_hash_state state;
   uint8_t text[INIT_SIZE_BYTE];
   uint8_t a[AES_BLOCK_SIZE];
-  uint8_t b[AES_BLOCK_SIZE];
+  uint8_t b[AES_BLOCK_SIZE * 2];
```
CodeIsTheKey, line 199: This was the root of the problem.  Depending on the compiler options, writing to b + AES_BLOCK_SIZE would write into c or a.  One was ok, the other killed the hash.

src/hash_selection.cpp
```diff
@@ -194,25 +194,17 @@ void coreHash(const void *toHash, uint512* hash, int lenToHash, int hashSelectio
...
-	  crypto::cryptonight_cnlite_hash(reinterpret_cast<const char*>(toHash), reinterpret_cast<char*>(hash), lenToHash, 1);
-		break;
-	 case 4:
-		crypto::cryptonight_turtle_hash(reinterpret_cast<const char*>(toHash), reinterpret_cast<char*>(hash), lenToHash, 1);
-		break;
-	 case 5:
-		crypto::cryptonight_turtlelite_hash(reinterpret_cast<const char*>(toHash), reinterpret_cast<char*>(hash), lenToHash, 1);
-		break;
-	}
+
```
CodeIsTheKey, line 197: No changes here, just cleanup for readability.

npq7721 (approved):
Great Job

Bigpiggy01:
Awesome!!!

PR #131, opened 2022-03-13
Fixed transaction time when scanning old blocks. Fixes #91

### LINKED ISSUE #91: Watched Addresses Have Incorrect Timestamp in Qt Wallet

CodeIsTheKey:
When scanning old blocks (e.g., when adding a watch address), the transaction time would be the time of the rescan, not the actual time of the transaction.

Reworked this PR for the current develop branch.

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -1109,7 +1109,7 @@ void CWallet::MarkDirty()
     fAnonymizableTallyCachedNonDenom = false;
 }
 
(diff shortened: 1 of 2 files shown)


npq7721 (approved):
LGTM

Issue #109, opened 2022-01-20
v0.16 v0.17 Cleanup - Qt Wallet


CodeIsTheKey:
Reviewing PRs #88 and #105, I noted the following files that need review/cleanup:

* Need to check out all of UI (terms, names, et. al.).
* src/qt/forms/receivecoinsdialog.ui - missing `<include location="../raptoreum.qrc"/>` - review.
* Search all qt translations - currently have Dash, not Raptoreum.
* src/qt/transactionrecord.cpp - redo refactor in pr #73

PR #134, opened 2022-03-19
Reworked transactions in Qt, rescan performance improvement

CodeIsTheKey:
Redid the transaction work in #73 
This also resolves the last issues for #109 

I found a performance issue when rescanning the wallet.  When I tried it on my wallet, a rescan from block 0 took 52 seconds.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -2096,7 +2096,22 @@ bool AppInitMain()
         ::feeEstimator.Read(est_filein);
     fFeeEstimatesInitialized = true;
 
-    // ********************************************************* Step 8: load wallet
+    // ********************************************************* Step 8a: load powcache.dat
+
+    {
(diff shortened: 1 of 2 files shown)


src/init.cpp
```diff
@@ -2096,7 +2096,22 @@ bool AppInitMain()
         ::feeEstimator.Read(est_filein);
     fFeeEstimatesInitialized = true;
 
-    // ********************************************************* Step 8: load wallet
+    // ********************************************************* Step 8a: load powcache.dat
```
CodeIsTheKey, line 2099: When the wallet was loaded (now in step 8b), it might trigger a rescan.  We want to have powcache loaded before that rescan occurs.

npq7721 (approved):
LGTM

PR #135, opened 2022-03-20
Minor test file cleanup

CodeIsTheKey:
Removed extra file (merge issue, combination rename then move).  It was renamed, then moved in two commits into the wallet tests, so it was lost during the merge.


npq7721 (approved):
LGTM

PR #138, opened 2022-03-24
testnet4

npq7721:
add setgenerate command
add future fee into future tx payload to validate when block is rescan
add future fee UI into send
add missing code from the dash 17 merged


src/evo/providertx.h
```diff
@@ -268,7 +269,8 @@ class CFutureTx {
         READWRITE(maturity);
 		READWRITE(lockTime);
 		READWRITE(lockOutputIndex);
-		READWRITE(updatableByDestination);
+        READWRITE(fee);
```
CodeIsTheKey, line 261: Check tabs in this file - looks like a mix.

src/Makefile.qt.include
```diff
@@ -193,6 +193,8 @@ BITCOIN_QT_H = \
   qt/winshutdownmonitor.h
 
 RES_ICONS = \
+  qt/res/icons/about.png \
```
CodeIsTheKey, line 196: Are these two images used?  I never found a reference.

src/qt/sendcoinsdialog.cpp
```diff
@@ -484,7 +516,9 @@ void SendCoinsDialog::accept()
 
 SendCoinsEntry *SendCoinsDialog::addEntry()
 {
-    SendCoinsEntry* entry = new SendCoinsEntry(this);
+
+    SendCoinsEntry* entry = new SendCoinsEntry(this, sporkManager.IsSporkActive(SPORK_22_SPEICAL_TX_FEE)
```
CodeIsTheKey: Typo - "SPEICAL"

CodeIsTheKey (approved):
LGTM - just a couple notes.

src/Makefile.qt.include
npq7721: it use in bitcoinui.cpp. i added it back

src/qt/sendcoinsdialog.cpp
npq7721: fixed

CodeIsTheKey:
Looks good!

Issue #157, opened 2022-04-09
Slow RPC call to getblockchaininfo during header sync


CodeIsTheKey:
### Describe the issue
Calling getblockchaininfo is useful to monitor a server.  During the initial header sync, this call can take a long time to respond (minutes).  This causes problems with monitoring scripts.

This is a problem with the current testnet release - 1.3.17.99.

PR #158, opened 2022-04-09
Slow RPC call to getblockchaininfo during header sync

CodeIsTheKey:
Resovles #157 
Added flag for header sync and atomic accessors for key blockchain data.

During header sync, the RPC call getblockchaininfo could take several minutes to return, causing problems with scripts monitoring nodes.  With the new logic in place, the RPC call was always returning quickly and during the entire sync process, the longest wait I had on my local machine was 157mSec.


npq7721 (approved):
Great Work

PR #160, opened 2022-04-12
Avoid potential deadlock when a new smart node is added.

CodeIsTheKey:
This fixes a potential deadlock when a new smart node is added.


npq7721 (approved):
tested and seem to work a lot better. thanks

PR #161, opened 2022-04-14
missing headers at src/chain.h

malbit:
Fixing compilation error under ubuntu due to:
"incomplete type std::atomic<size_t> atomicHeight"
included and brought at #158

### DIFF
--- a/src/chain.h
+++ b/src/chain.h
@@ -6,6 +6,9 @@
 #ifndef BITCOIN_CHAIN_H
 #define BITCOIN_CHAIN_H
 
+#include <atomic>
+#include <memory>
+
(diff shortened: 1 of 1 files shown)


CodeIsTheKey:
We definitely need to include atomics.  Why is `#include <memory>` needed?

malbit:
> We definitely need to include atomics. Why is `#include <memory>` needed?

<memory> is used for std::unique_ptr basically. 
If I did add it I might see some unique pointers in a file.
Double check if any do exist

CodeIsTheKey (approved):
LGTM - thanks for catching that!

npq7721 (approved):

PR #162, opened 2022-04-24
Powcache save

CodeIsTheKey:
Added logic to periodically save the POW cache.  This saves a lot of restart time for users if the server is shutdown suddenly.


npq7721 (approved):
Great work. thanks

PR #141, opened 2022-03-25
Raptoreum v17 with massive backports addoption from bitcoin and dash

malbit:
- Update crc32c subtree (ref: https://github.com/bitcoin/bitcoin/commit/fac1c13ead8da2414acdd3202040f00a21435b8f)
- Update univalue subtree (ref: https://github.com/bitcoin/bitcoin/commit/304319367595b51abfd69f1c4abddeef0acca3a9)
- Update leveldb subtree (ref: https://github.com/bitcoin/bitcoin/commit/20a6babfa9a66f5432ef19c6c433b4357560f853)
- refactor: depends build env
- refactor: build configuration (comply with autotools 2.69)
- Integrate univalue into our build system (ref: https://github.com/bitcoin/bitcoin/commit/0f95247246344510c9a51810c14c633abb382e95)
- crc32c build system integration (ref: https://github.com/bitcoin/bitcoin/commit/9ebdf047578f0da7e6578d0c51c32f55e84ac157)
- fixed placeholders at QT
- Support to checking HW functions at compile-time
- Support for HW 256-bit RNG (where available)

PR #164, opened 2022-05-08
Fix libbacktrace compile on alpine/musl arch.

michal-zurkowski:
Alpine is using musl architecture. It is not properly detected now with the 2018 version of libbacktrace.
This fix allows proper dependency build on the alpine system.
I tested it on the current develop chain and it compiled, synced, and run just fine.

Same thing as was referenced and pushed on Dash repo: https://github.com/dashpay/dash/pull/4417


chadouming:
+1, that was also my fix. This is also part of the pull request #141 .

npq7721 (approved):
Thanks

Issue #154, opened 2022-04-04
Weird bug with QT wallet and futures


sullystuff:
To replicate: 
Start creating a transaction with GUI wallet, select a few inputs. Enable future, and set values to whatever (ex. 100 and 100)
Set output such that it uses most of the inputs and does not leave enough for the future transaction.
Try to send tx, press cancel.
Disable future. 
Try to send tx again. Future will still be enabled.

I haven't done extensive testing so some of those steps may or may not be required - that's just what led me to the issue.
![image](https://user-images.githubusercontent.com/63869300/161618729-83705f8f-c70e-4e03-a70f-27ea47963269.png)

PR #166, opened 2022-07-06
Enable/disable future transaction

nandofw:
fix for issue  [Weird bug with QT wallet and futures #154 ](https://github.com/Raptor3um/raptoreum/issues/154)


CodeIsTheKey:
Makes sense - this is the only place isFutureOutput is assigned.  LGTM

npq7721 (approved):
good catch. thanks

PR #167, opened 2022-07-09
Use atomic instead of lock for RPC getblockcount.

CodeIsTheKey:
Updated RPC getblockcount call to use atomic fetch instead of locked access to prevent delays.


npq7721 (approved):
LGTM

PR #170, opened 2022-07-30
Clarified "Subtract fee from amount" when sending future transactions

### LINKED ISSUE #150: "Subtract fee from amount" does not subtract 100 RTM when futures transaction
When sending a futures transaction with the GUI wallet, using 'subtract fee from amount' will only subtract the standard network fee, and not the 100 RTM futures fee.
image

CodeIsTheKey:
I looked at subtracting the future tx fee from the amount, but realized that it is handled separately from the normal network fees.

Instead of changing the model to track two sets of fees, I have updated the checkbox on the dialog to indicate network fees, and it now shows the networks fees and special transaction fee in the confirmation dialog.

### Single transaction:
image
image

### Multiple transactions:
image
image

This resolves #150

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Raptoreum Core staging tree 0.17
 |CI|master|develop|
 |-|-|-|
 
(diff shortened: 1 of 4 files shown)


CodeIsTheKey:
Closing - #171 works well and is a better solution.

Issue #40, opened 2021-11-09
No icons for future transaction when pending in transactions tab


Raptor3um:
When future is received it has no icon until confirmed 6 times. It should have an hourglass (screen1) until first confirmation, after that turns to loading type icon (screen2).

![future receive no icon](https://user-images.githubusercontent.com/47279850/140944101-5b5c0e11-5252-41a9-b443-38c98aba1f0a.png)
![future receive no icon2](https://user-images.githubusercontent.com/47279850/140944100-046fac13-055e-4389-82d0-16f810c67603.png)

PR #172, opened 2022-08-02
Add icon future transaction

nandofw:
Add icon for future transaction. should solve #40 
fix incorrect selection of future transaction maturity type. always showed as based on time
fix future transaction show as confirmed while still on mempool

image
image
image


npq7721 (approved):
Good work. thanks

PR #173, opened 2022-08-05
update macos version action build to 11

npq7721:
macos version 10.15 no longer work for github actions so updated it to version 11


nandofw (approved):
executable works well on macOS 10.13. LGTM

Bigpiggy01 (approved):
Let's go!!!

PR #175, opened 2022-08-10
update mainnet parameters and move dash release docs to dash folder

npq7721:


src/util.h
```diff
@@ -59,7 +59,8 @@ extern int nWalletBackups;
 // Application startup time (used for uptime calculation)
 int64_t GetStartupTime();
 
-static const int DEFAULT_POW_CACHE_SIZE = 150000;
+static const int DEFAULT_POW_CACHE_SIZE = 300000;
```
CodeIsTheKey: Recommend raising this to 1000000.  powcache.dat is small compared to the blocks database, so it is an efficient improvement.

CodeIsTheKey (approved):
Looks good to me!  One suggested change.

src/util.h
npq7721: changed as recommanded

PR #178, opened 2022-08-16
Fix scale on recent transactions

### LINKED ISSUE #147: "Recent Transactions" doesn't scale correctly
After adjusting "Font Scale" closer to "Smaller" in Appearance section of Options, the text in the "Recent transactions" section of the "Overview" tab of the wallet gets much smaller than the rest of the text in the wallet.

Before: 
image

After:
image

nandofw:
Removed duplicated call to getScaledFontSize

Should fix #147


CodeIsTheKey:
I built and looked at the result - they look good to me!

CodeIsTheKey (approved):

PR #179, opened 2022-08-17
minor UI improvements and bug fixed

npq7721:
fixed type4 llmq on testnet
increase coinjoin max limit to 2M with increment of 1000
increase hard fork block to 410000


CodeIsTheKey (approved):
Looks good.

PR #182, opened 2022-08-17
Fixed bug causing unnecessary transaction refreshes

CodeIsTheKey:
Tracked down a performance issue where every new block would trigger a complete refresh of the transactions in the wallet(s).  I was testing with a wallet with 127K transactions and the UI would hang for about 90 seconds every block.  This was because it thought it needed to refresh the transaction when the block height changed - even for confirmed transactions.

With the update, the initial update is still required, but subsequent updates are very quick.


npq7721 (approved):
Excellent improvement

PR #183, opened 2022-08-18
Fixed logic in Qt transaction list when watched address is one of many outputs

CodeIsTheKey:
In transactions where there are large distributions of coins (e.g., mining payments) and the watched addresses is one of the recipients, the Qt wallet would show every output in the transaction as a watched transaction (usually 100s of transactions).  I corrected this so that is only shows the vout for the watched address(es).


npq7721 (approved):
LGTM

PR #184, opened 2022-08-21
Prevent future option begin displayed early

nandofw:
prevent future option begin displayed if SPORK_22_SPECIAL_TX_FEE is active while block count is less that consensus.nFutureForkBlock

removed leftover unused/dead code


CodeIsTheKey (approved):
Looks good.  Thanks for fixing this!

npq7721 (approved):
LGTM

PR #187, opened 2022-08-24
change MAX_MONEY to 21b

nandofw:
In order to change MAX_MONEY from 21 milion to 21 bilion and not break actual consensus during the transition period
-qt and cli wallet are limited to 21 million
-ProcessNewBlock and mempool will not accept any transaction over 21m

All changes were tested on devnet and seems to work properly. Logs are available if requested

Issue:
coinjoin is still limited to 21m


src/consensus/tx_verify.cpp
```diff
@@ -271,7 +277,7 @@ bool CheckTransaction(const CTransaction& tx, CValidationState &state, int nHeig
     return true;
 }
 
-bool Consensus::CheckTxInputs(const CTransaction& tx, CValidationState& state, const CCoinsViewCache& inputs, int nSpendHeight, CAmount& txfee, CAmount& specialTxFee, bool fFeeVerify)
+bool Consensus::CheckTxInputs(const CTransaction& tx, CValidationState& state, const CCoinsViewCache& inputs, int nSpendHeight, CAmount& txfee, CAmount& specialTxFee, bool isV17active, bool fFeeVerify)
```
npq7721, line 280: isV17active passing inhere is not necessary since u can do bool isV17active = Params().IsFutureActive(chainActive.Tip());

npq7721 (changes requested):
check my comment

src/consensus/tx_verify.cpp
nandofw: Consensus::CheckTxInputs fuction don't have access to Params().IsFutureActive(chainActive.Tip()). that's why i'am passing the isV17active to it

npq7721: you use it in line 228 of the same file that CheckTxInputs function is https://github.com/Raptor3um/raptoreum/blob/3cc119cca2465b5a6770cbbf84e1be2181a1272f/src/consensus/tx_verify.cpp#L228 so it should work

nandofw: this function is under Consensus namespace and if i  try to access ChainParams data there it fail to compile
image

npq7721 (approved):

PR #189, opened 2022-08-29
Fixed handling and documentation for blocksdir.

### LINKED ISSUE #181: Extra empty directory created when doing full sync
This issue with the current develop branch.

When I start a fresh sync (on testnet4), it is creating a `blocks` directory in the wrong location.  It then proceeds to create the blocks directory in the correct location and using it.

Review the startup sequence and do not create this extraneous directory.

When I was running it, I specified a data directory - I don't know if this makes a difference or not.

Here is how I started raptoreum-qt (I assume the issue is the same for raptoreumd):

`raptoreum-qt --datadir=/home/codeiskey/rtm/testnet`

This creates:
`/home/codeiskey/rtm/testnet/blocks` which is empty

And it creates:
`/home/codeiskey/rtm/testnet/testnet4/blocks` which is used.

CodeIsTheKey:
Fixes #181 

Pulled in from Bitcoin:
https://github.com/bitcoin/bitcoin/pull/14364
https://github.com/bitcoin/bitcoin/pull/14409

Tested with mainnet and testnet4.  If the user specifies a --blocksdir, they are required to create that directory.

Here is the directory structure after specifying --datadir=testnet and --blocksdir=testnet/BLOCKSDIR (they were absolute paths, shortened for convenience):
image

Here is the same run without --blocksdir:
image

Here is mainnet with --datadir and --blocksdir set (I manually created BLOCKSDIR):
image

Here is mainnet with only --datadir:
image


npq7721 (approved):
LGTM

PR #190, opened 2022-08-29
Qt: UX improvement on POW cache first time initialization.

jakegny:
If the `powcache.dat` file is not present, show a user message while that is recreated / initialized.


npq7721 (approved):
LGTM

PR #191, opened 2022-09-06
Fix GUI freezing while sync blocks

nandofw:


npq7721 (approved):
Exact what i have in mind. excellent job

PR #192, opened 2022-09-09
Fixed bug when smartnode list changes during request

CodeIsTheKey:
During certain time sensitive updates, retrieving the smartnode list could change during the call, possibly causing a crash.  Protected from the possible crash.


CodeIsTheKey:
The crash can occur if the result is resized to *larger* than its actual size.  Some callers retrieve the last element of the vector, which is a null pointer, and dereference it, causing the crash.

npq7721 (approved):
LGTM

PR #195, opened 2022-09-09
Large wallet performance improvements

CodeIsTheKey:
Performance improvements for large wallets (particularly transactions with many inputs or outputs).  Added caching of WalletTx and used shared_ptr to minimize excessive copying.  Performance for this logic for smaller wallets is ~20x faster, for larger wallets, well over 100x faster.

### DIFF
--- a/src/interfaces/wallet.cpp
+++ b/src/interfaces/wallet.cpp
@@ -56,41 +56,41 @@ public:
 };
 
 //! Construct wallet tx struct.
-WalletTx MakeWalletTx(CWallet& wallet, const CWalletTx& wtx)
+shared_ptr<WalletTx> MakeWalletTx(CWallet& wallet, const CWalletTx& wtx)
 {
-    WalletTx result;
+    shared_ptr<WalletTx> result(new WalletTx);
     bool fInputDenomFound{false}, fOutputDenomFound{false};
-    result.tx = wtx.tx;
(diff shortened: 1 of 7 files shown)


src/interfaces/wallet.cpp
```diff
@@ -173,6 +173,7 @@ class CoinJoinImpl : public CoinJoin::Client
 class WalletImpl : public Wallet
 {
 public:
+    unordered_lru_cache<uint256, shared_ptr<WalletTx>, std::hash<uint256>, 10000> m_WalletTxCache;
```
CodeIsTheKey, line 176: Creates the cache of WalletTx.  These were being recreated continuously and they don't change.  With a lot of vins/vouts, it was rebuilding vectors every call and all of this too a lot of time.  Additionally, with many vouts this creation was being done for each vout for the same Tx.

src/qt/transactionrecord.cpp
```diff
@@ -18,6 +18,7 @@
 #include <utilmoneystr.h>
 
 #include <stdint.h>
+#include <QDebug>
```
npq7721: what is this use for?

CodeIsTheKey: Fixed - this was left over from when I had debug logging.

npq7721 (approved):
LGTM

PR #202, opened 2022-09-11
emission curve

nandofw:
correction on the emission curve


CodeIsTheKey:
Trouble spots before change:
image

After change:
image

CodeIsTheKey (approved):
Looks good.

PR #208, opened 2022-09-15
attempt to fix smartnode

nandofw:
As PROTOCOL_VERSION has changed it conflict with old version as the contribution encrypt/decrypt  function receive PROTOCOL_VERSION as parameter


npq7721 (approved):
Good catch thanks

PR #215, opened 2022-09-25
fix devnet, regtest and update test

nandofw:
-fix for devnet and regtest
-Updated some test units


CodeIsTheKey (approved):
Looks great!  I tested both regtest and devnet and they are working well.  There are some very long running unit tests that we should improve (different isue), but the tests run much better.

npq7721 (approved):

PR #217, opened 2022-09-30
Disable unnecessary warning

Leonardiae:
While syncing my node from scratch this warning first scared me:
`IsBlockPayeeValid -- WARNING: Client synced but old budget system is disabled, accepting any payee`
Then I read on discord that someone else had the same question before and Bigpiggy answered:  "just means were are not running the dash system for budget" 

So I figured why not disable this warning, since WARNING sounds kind of serious.

Maybe this change will help prevents others having the same question about this warning in the future.


Bigpiggy01:
Hi man this should probably be up as an issue not as a pull request unless you plan to add code to it?

npq7721 (commented):
i would not remove the comment completely, would just put log on different level so normally u don't see it. so if we decide to have this feature we may be using this log for debugging

Leonardiae:
@npq7721 after your suggestion I changed the loglevel. I have not much experience with pull request so please bear with me. Just trying to help and learn :)

npq7721 (approved):
LGTM

npq7721:
@Leonardiae can redo the PR on develop branch i will revert this commit on master, PR should only do to develop

PR #219, opened 2022-10-04
Change loglevel of warning

Leonardiae:
While syncing my node from scratch this warning first scared me: 
`IsBlockPayeeValid -- WARNING: Client synced but old budget system is disabled, accepting any payee` 

Then I read on Discord that someone else had the same question before and Bigpiggy answered: "just means were are not running the dash system for budget" So I figured why not change the loglevel to debug, maybe this change will help prevent others having the same question about this warning in the future.

This pr is a copy of https://github.com/Raptor3um/raptoreum/pull/217#event-7511008469 which I wrongly did on master instead of develop.


npq7721 (approved):
LGTM

PR #225, opened 2022-10-08
Fixed script standard tests

nandofw:
Solver return false on RTM for TX_NONSTANDARD while on on bitcoin it return true


CodeIsTheKey (approved):
Nice - looks good!

PR #226, opened 2022-10-09
Fixed txvalidationcache tests

nandofw:
txvalidationcache_tests require BIP66 to be off in order to run (default is on).
Added logic to turn on/off BIP66, usage only allowed on regtest


CodeIsTheKey (changes requested):
This all looks good and the change makes sense, I only see one issue.
Please change `nactive` to `active` - the leading `n` implies this argument is an integer.

Thanks!

nandofw:
> This all looks good and the change makes sense, I only see one issue. Please change `nactive` to `active` - the leading `n` implies this argument is an integer.
> 
> Thanks!

Done changes applied

CodeIsTheKey (approved):

npq7721 (approved):

PR #227, opened 2022-10-09
Updated miner_tests

CodeIsTheKey:
* Rebuilt nonces for RTM mainnet
* Updated test amount due to small 4 RTM fair launch block rewards
* Updated logic to support DIP3
* Corrected to test for 40,000 max sigops per block
* Updated to properly handle BIP68 active at start


nandofw (approved):
LGTM

PR #228, opened 2022-10-10
Fixed coins tests

nandofw:
Fixed serialization error on coins_tests


CodeIsTheKey (approved):
Looks good and works locally for me.  It hhandles the two new fields that have been added.  After these tests are all running, we should add a test for futureTx serialization/deserialization.

Issue #229, opened 2022-10-10
No-wallet compiling includes issues


ghost:
### Describe the issue
Compiling the master branch with --disable-wallet fails due to the following error: 
```
rpc/rawtransaction.cpp:557:9: error: 'UpdateSpecialTxInputsHash' was not declared in this scope
  557 |         UpdateSpecialTxInputsHash(rawTx, ftx);
```

My guess is that this is either due to the no-wallet compile wrongfully requiring this dependency in rawtransaction.cpp or UpdateSpecialTxInputsHash wrongfully being within #ifdef ENABLE_WALLET tags.

### Can you reliably reproduce the issue?

Not outside of my current environment. I'm using FreeBSD and have made very many modifications already to get this far in the compilation process. This doesn't mean this doesn't happen on other systems as well I just haven't tried.
I will most likely reset my vps after submitting this issue because I'll probably just use linux binary compatibility instead of building my own.

### What version of Raptoreum Core are you using?
The master branch source code of:
[1.3.17.01](https://github.com/Raptor3um/raptoreum/releases/tag/1.3.17.01)

### Machine specs:
- OS: FreeBSD 13.1-RELEASE
- CPU: 4 virtual cores (AMD EPYC 7282 16-Core Processor)
- RAM: 8gb
- Disk size: 50gb+ free space
- Disk Type (HD/SDD): NVMe

### Any extra information that might be useful in the debugging process.

These are the commands I use to build:
```
./autogen.sh
./configure --disable-tests --disable-gui-tests --disable-dependency-tracking \
--disable-bench --with-gui=no --without-miniupnpc --without-qtdbus \
--enable-miner=no --disable-wallet \
MAKE=gmake BDB_FLAGS="-I/usr/local/include/db4"
```

It doesn't contain anything related to the depends folder due to none of that working on FreeBSD. I have figured out and installed everything (as far as I can tell) required manually.

PR #230, opened 2022-10-10
fix compiling with --disable-wallet and --enable-miner=no

nandofw:
-Compile with --disable-wallet give the following error
`rpc/rawtransaction.cpp:557:9: error: 'UpdateSpecialTxInputsHash' was not declared in this scope
  557 |         UpdateSpecialTxInputsHash(rawTx, ftx);`
  
 -Compile with --enable-miner=no give the following error
`/miner.cpp:547: undefined reference to 'EnsureWalletIsAvailable(CWallet*, bool)'`

This pr fix both issue and should solve #229

changes tested under Ubuntu 20.04

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -543,10 +543,11 @@ void static RaptoreumMiner(const CChainParams& chainparams)
 
     #ifdef ENABLE_WALLET
         pWallet = GetFirstWallet();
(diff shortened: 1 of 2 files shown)


npq7721 (approved):
LGTM

PR #235, opened 2022-10-15
Fixed dip3 tests

nandofw:
Updated test amount due to small 4 RTM fair launch block rewards
Increased the number of smartnode created in order to start payment


npq7721 (approved):
LGTM

PR #250, opened 2022-11-20
[Assets] Assets creation transaction

nandofw:
- added new type of transaction: TRANSACTION_NEW_ASSET 
- implemented CNewAssetTx containing asset metadata
- added nAssetsForkBlock on chainparams
- added unit test that create a new assets

### DIFF
--- a/src/evo/providertx.h
+++ b/src/evo/providertx.h
@@ -291,8 +291,91 @@ public:
     }
 };
 
+class CNewAssetTx {
+public:
+    static const uint16_t CURRENT_VERSION = 1;
+
+    uint16_t nVersion{CURRENT_VERSION};// message version
+    std::string Name;
+    bool updatable = true;//if true this asset meta can be modify using assetTx update process. 
+    bool isunique = false;//true if this is asset is unique it has an identity per token (NFT flag)
+    uint8_t Decimalpoint = 0;
+    std::string referenceHash; //hash of the underlying physical or digital assets, IPFS hash can be used here.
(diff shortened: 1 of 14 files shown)


src/evo/providertx.h
```diff
@@ -291,9 +291,92 @@ class CFutureTx {
...
 };
 
+class CNewAssetTx {
+public:
+    static const uint16_t CURRENT_VERSION = 1;
+
+    uint16_t nVersion{CURRENT_VERSION};// message version
+    std::string Name;
+    bool updatable = true;//if true this asset meta can be modify using assetTx update process. 
+    bool isunique = false;//true if this is asset is unique it has an identity per token (NFT flag)
```
npq7721: should be isUnique

src/evo/providertx.h
```diff
@@ -291,9 +291,92 @@ class CFutureTx {
...
+    uint8_t Decimalpoint = 0;
```
npq7721: should be decimalPoint

npq7721 (changes requested):
unlike future transaction we want create new asset transactions to follow the same logic as registering protx where it get store in a merkle root proof in mined block. we can create another issue to add those if it is end up taking too big for this PR

src/evo/providertx.h
nandofw: done

nandofw:
- added asset metadata validation

npq7721 (approved):
LGTM

PR #251, opened 2022-11-23
Bring "Options" item under "Settings" menu on Mac

rocketsvm:
Should make docs and tutorials more consistent between systems. I'm not a Mac user, but there were people having trouble finding this item e.g. for enabling coin control features.


npq7721:
please do PR into develop branch instead of master

PR #254, opened 2022-12-02
fix sync issue

nandofw:
Fixed synchronization issue in block 453100 caused by a quorum member list error


CodeIsTheKey (commented):
I verified full sync with this logic in place without issue.  453100/453101 worked well and the proper node was PoSE banned afterwards.

npq7721 (approved):
great job

PR #256, opened 2022-12-15
update fix for block 453100

nandofw:
-added collateral amount on CDeterministicMNState this is used to check if the collateral is valid at a determined block height
-reverted #254 as is no longer necessary

this pr require a sync from 0

### DIFF
--- a/src/evo/deterministicmns.h
+++ b/src/evo/deterministicmns.h
@@ -48,6 +48,8 @@ public:
     // sha256(proTxHash, confirmedHash) to speed up quorum calculations
     // please note that this is NOT a double-sha256 hash
     uint256 confirmedHashWithProRegTxHash;
+    //collateral amount used in quorum calculations
+    CAmount ncollateralamount;
 
     CKeyID keyIDOwner;
     CBLSLazyPublicKey pubKeyOperator;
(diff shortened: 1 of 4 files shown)


npq7721 (approved):
good solution, LGTM

nandofw:
fixed tests error when collateral is a output of the TRANSACTION_PROVIDER_REGISTER

src/evo/deterministicmns.h
```diff
@@ -48,6 +48,8 @@ class CDeterministicMNState
     // sha256(proTxHash, confirmedHash) to speed up quorum calculations
     // please note that this is NOT a double-sha256 hash
     uint256 confirmedHashWithProRegTxHash;
+    //collateral amount used in quorum calculations
+    CAmount ncollateralamount;
```
CodeIsTheKey: Suggest changing case to match other fields: `nCollateralAmount`

src/evo/deterministicmns.h
```diff
@@ -154,6 +157,7 @@ class CDeterministicMNStateDiff
         Field_addr                              = 0x0800,
         Field_scriptPayout                      = 0x1000,
         Field_scriptOperatorPayout              = 0x2000,
+        Field_ncollateralamount                 = 0x4000,
```
CodeIsTheKey: Suggest `Field_nCollateralAmount`

CodeIsTheKey (approved):
Great solution!  Nice way to avoid the index and get a good collateral check.

src/evo/deterministicmns.h
nandofw: done

PR #258, opened 2022-12-30
Add minrefresh option

CodeIsTheKey:
I had added an option (`minrefresh=1`) for the Qt wallet to minimize UI refreshes every transaction after each new block.  For wallets with a large number of transactions, this refresh can be problematic.  This refresh is bypassed only for fully confirmed transactions.

The default (`minrefresh=0`) is to refresh after each block, so the confirmation count will rise as expected.

I also added block height to the transaction details in the tooltip.  I find myself looking for that regularly.


npq7721 (approved):
LGTM

PR #259, opened 2023-01-11
[Assets] Asset update transaction

nandofw:
- added new type of transaction: TRANSACTION_UPDATE_ASSET
- implemented CUpdateAssetTx containing asset update metadata

tests will made later as it require cache implementation


npq7721 (approved):
LGTM

PR #263, opened 2023-01-19
[Assets] opcode OP_ASSET_ID

nandofw:
- introduction of new opcode OP_ASSET_ID
- introduced a prototype for asset transaction script
- included a basic test case for asset transaction script

### DIFF
--- a/src/script/script.cpp
+++ b/src/script/script.cpp
@@ -6,6 +6,7 @@
 #include <script/script.h>
 #include <tinyformat.h>
 #include <utilstrencodings.h>
+#include <assets/assets.h>
 
 const char* GetOpName(opcodetype opcode)
 {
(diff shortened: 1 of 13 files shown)


src/script/script.cpp
```diff
@@ -143,13 +144,44 @@ const char* GetOpName(opcodetype opcode)
...
+
+bool CScript::IsAssetScript(int& nIndex) const
+{
+    if (this->size() > 31) {  
+        if ((*this)[25] == OP_ASSET_ID) { // OP_RTM_ASSET is always in the 25 index of the script if it exists
+            nIndex = -1;
+            if ((*this)[27] == RTM_R) { // Check to see if RTM starts at 27 ( this->size() < 105)
+                if ((*this)[28] == RTM_T)
+                    if ((*this)[29] == RTM_M)
+                        nIndex = 30;
```
npq7721, line 170: what is the idea of RTM asset OP code size can be < 105 or >= 105?

nandofw: this is because the serialization of vchMessage on  BuildAssetTransaction add 1 or 2 byte to the script depending on its size

npq7721 (approved):
LGTM

PR #264, opened 2023-01-20
Add pose penalty to smartnode list

CodeIsTheKey:
Added posepenalty to the RPC `smartnode list` call.  I find it very useful to monitor the network and quorums.

For example, you can get the sum of all PoSe penaltieis like this (the Network PoSe Penalty):
`raptoreum-cli smartnode list | grep -i posepenalty | awk '{ sum += $2 } END { print sum }'`


nandofw (approved):
LGTM

npq7721 (approved):

PR #266, opened 2023-02-15
add spendableHeight and spendableTime to getaddressutxos output

npq7721:
they were missing from getaddressutxos so client currently would not able to determine whether an unspent is being locked


CodeIsTheKey (approved):
LGTM!

PR #271, opened 2023-02-18
Auto lock smartnode collaterals after sync completes.

CodeIsTheKey:
When performing a full sync, smartnode collateral transactions are not locked when the sync completes.  The problem is that during initialization, those transactions are locked - but we have not synced the transactions, yet, so they are not locked.

I have added a one-time re-lock of collateral transactions when the sync completes.


CodeIsTheKey:
Fixes #272

npq7721 (approved):
LGTM

PR #276, opened 2023-02-25
[Assets] database cache,  Wallet and RPC integration

nandofw:
- added a basic database cache for assets
- added wallet support for assets
- added asset transaction signing
- adjusted OP_ASSET_ID to fix a failing test
- added consensus rules for asset transactions
- added createasset, mintasset, sendasset, assetdetails and listassetsbalance RPC call

### DIFF
--- a/src/consensus/tx_verify.cpp
+++ b/src/consensus/tx_verify.cpp
@@ -348,6 +348,32 @@ bool CheckTransaction(const CTransaction& tx, CValidationState &state, int nHeig
     return true;
 }
 
+inline bool checkOutput(const CTxOut& out, CValidationState& state, CAmount& nValueIn, std::map<std::string, CAmount>& nAssetVin, bool isV17active)
+{
+    // Check for negative or overflow values
+    nValueIn += out.nValue;
+    if (!MoneyRange(out.nValue, isV17active) || !MoneyRange(nValueIn, isV17active)) {
+        return state.DoS(100, false, REJECT_INVALID, "bad-txns-values-outofrange");
+    }
+
+    if(out.scriptPubKey.IsAssetScript()){
+        CAssetTransfer assetTransfer;
+        if(!GetTransferAsset(out.scriptPubKey, assetTransfer)){
+            return state.DoS(100, false, REJECT_INVALID, "bad-assets-input");
+        }
+        if(nAssetVin.count(assetTransfer.AssetId))
+            nAssetVin[assetTransfer.AssetId] += assetTransfer.nAmount;
+        else
+            nAssetVin.insert(std::make_pair(assetTransfer.AssetId, assetTransfer.nAmount));
+        
+        if (!MoneyRange(assetTransfer.nAmount) || !MoneyRange(nAssetVin.at(assetTransfer.AssetId))) {
+            return state.DoS(100, false, REJECT_INVALID, "bad-asset-values-outofrange");
+        }
+    }
+
+    return true;
+}
+
 bool Consensus::CheckTxInputs(const CTransaction& tx, CValidationState& state, const CCoinsViewCache& inputs, int nSpendHeight, CAmount& txfee, CAmount& specialTxFee, bool isV17active, bool fFeeVerify)
 {
     // are the actual inputs available?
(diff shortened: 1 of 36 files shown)


src/consensus/tx_verify.cpp
```diff
@@ -289,6 +291,27 @@ bool CheckTransaction(const CTransaction& tx, CValidationState &state, int nHeig
         nValueOut += txout.nValue;
         if (!MoneyRange(nValueOut, isV17active))
             return state.DoS(100, false, REJECT_INVALID, "bad-txns-txouttotal-toolarge");
+
```
npq7721, line 294: can we move these assets related check into its own function so it a bit easier to follow? this function is getting longer as time goes one.

src/consensus/tx_verify.cpp
```diff
@@ -351,14 +375,59 @@ bool Consensus::CheckTxInputs(const CTransaction& tx, CValidationState& state, c
             return state.DoS(100, false, REJECT_INVALID, "bad-txns-inputvalues-outofrange");
         }
 
+        if(coin.out.scriptPubKey.IsAssetScript()){
```
npq7721: move this to different function as well, and it seem that this a repeated pattern so i think better to share the same function. i understand it has different error message but the pattern almost identical

src/consensus/tx_verify.cpp
```diff
@@ -351,14 +375,59 @@ bool Consensus::CheckTxInputs(const CTransaction& tx, CValidationState& state, c
...
 
-    const CAmount value_out = tx.GetValueOut();
+    CAmount value_out = 0;
+    std::map<std::string, CAmount> nAssetVout;
+    for (auto out : tx.vout){
+        value_out += out.nValue;
+        if (!MoneyRange(out.nValue, isV17active) || !MoneyRange(value_out, isV17active)) {
+            return state.DoS(100, false, REJECT_INVALID, "bad-txns-outputvalues-outofrange");
+        }
+        if(out.scriptPubKey.IsAssetScript()){
```
npq7721: same here as above comments

npq7721 (changes requested):

src/consensus/tx_verify.cpp
nandofw: done. moved this repeated code to a single function to handle inputs/outputs checks

nandofw:
Updated creation/mint logic for unique asset

src/consensus/tx_verify.cpp
```diff
@@ -325,6 +348,32 @@ bool CheckTransaction(const CTransaction& tx, CValidationState &state, int nHeig
...
+inline bool checkOutput(const CTxOut& out, CValidationState& state, CAmount& nValueIn, std::map<std::string, CAmount>& nAssetVin, bool isV17active)
+{
+    // Check for negative or overflow values
+    nValueIn += out.nValue;
+    if (!MoneyRange(out.nValue, isV17active) || !MoneyRange(nValueIn, isV17active)) {
+        return state.DoS(100, false, REJECT_INVALID, "bad-txns-values-outofrange");
+    }
+
+    if(out.scriptPubKey.IsAssetScript()){
+        CAssetTransfer assetTransfer;
```
npq7721, line 363: need to check somewhere in here that if this is asset script of any kind, out.nValue should always be 0

src/evo/providertx.cpp
```diff
@@ -108,7 +109,25 @@ bool CheckFutureTx(const CTransaction& tx, const CBlockIndex* pindexPrev, CValid
...
+{
+    if (assettx.decimalPoint > 0 ){ // alway 0
+        return state.DoS(100, false, REJECT_INVALID, "bad-assets-decimalPoint"); 
+    }
+    
+    if (assettx.type > 0 ){ // manual mint only?
+        return state.DoS(100, false, REJECT_INVALID, "bad-unique-assets-distibution-type");
+    }
+
+    if (assettx.Amount != 1 * COIN){ //unique suply = 1 COIN
```
npq7721, line 123: an unique asset does not mean supply is always 1. unique asset mean each individual asset coin tie to an identity so each asset coin may not created equals. these indentation (id) that associated with each coin is specified in asset output script. for example, let say a unique asset token name Taylor_Swift_concert_ticket (TSCT for short), it first mint transaction produce 100 TSCT, the mint transaction would have asset amount = 100 and the output script would contain id=[1-100]. we want to use range here as much as possible to avoid listing 1,2,3...., 100. now user transfer 1 TSCT to another address, they can specify which id they sending. and the change output would have the remaining id and amounts. In this example TSCT[1] would represent a front row seat at the concert and would be more example then TSCT[100] which let it is back row. 

This change may be a lot so i suggest we approve this PR with other changes request above and make another PR for this unique asset.

npq7721 (changes requested):

src/consensus/tx_verify.cpp
nandofw: fixed

npq7721 (approved):

PR #281, opened 2023-03-05
Fix vout shuffling in FundTransaction.

michal-zurkowski:
This should fix the reordering of vout transactions while using FundTransaction.

The current code can reorder vout amount that is sent while keeping not corresponding script/address
example transaction (WALLET: AMOUNT)
X: 20.0
Y: 30.0
Z: 10.0
After FundTransaction (fundrawtransaction in a wallet)
X: 10.0
Y: 20.0
Z: 30.0

Amounts are sorted according to BIP69 but the addresses are not changed with it right now.
This results in a completely broken send which can go through, but the recipients will get completely different amounts in most cases.

New code fixes that issue.


nandofw (approved):
LGTM

PR #283, opened 2023-03-19
[Assets] unique assets, consensus, qt send asset

nandofw:
- adjusted consensus checks for unique assets
- added mempool checks to prevent duplicate assets
- added mempool checks for duplicates mint tx, this is necessary to prevent duplicates uniqueId on unique assets (NFT)
- added more test cases

[QT]
- added Send Asset page
- added Asset control features

Send asset:
image
Send unique asset:
image
Asset control:
image

### DIFF
--- a/src/Makefile.qt.include
+++ b/src/Makefile.qt.include
@@ -36,6 +36,7 @@ QT_FORMS_UI = \
   qt/forms/appearancewidget.ui \
   qt/forms/askpassphrasedialog.ui \
   qt/forms/coincontroldialog.ui \
+  qt/forms/assetcontroldialog.ui \
   qt/forms/editaddressdialog.ui \
   qt/forms/helpmessagedialog.ui \
   qt/forms/intro.ui \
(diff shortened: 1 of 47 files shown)


CodeIsTheKey (commented):
In assets.h, CAssetOutputEntry is a struct (probably should be a class) and forward defined. The definition is in wallet/wallet.h - is this a problem when we split out the wallets from core?

nandofw:
> In assets.h, CAssetOutputEntry is a struct (probably should be a class) and forward defined. The definition is in wallet/wallet.h - is this a problem when we split out the wallets from core?

Good point, CAssetOutputEntry is only used by the wallet so the compiler should handle it. well if there is any issue the GetAssetData function can be moved to wallet/wallet.h

CodeIsTheKey (approved):
I focused on the core code (not qt).  The tests and logic look good to me.

I know Tri wants to review the architecture vs design, but this looks good to me!

npq7721 (approved):

PR #287, opened 2023-04-23
Added ability to manually trigger a build

CodeIsTheKey:
This will let us manually trigger builds on branches when needed.


npq7721 (approved):
LGTM

PR #288, opened 2023-04-23
avoid updating the balances unnecessarily

nandofw:
while watching flockpool fees address that has a large amount of transaction, there is a constant cpu usage of 8-9%, with this changes the cpu usage dropped to almost 0%


CodeIsTheKey (approved):
LGTM - nice catch!

PR #303, opened 2023-07-11
assets_mint test fix

CodeIsTheKey:
The assets_mint test had a temporary workaround for a test failure under Windows.  The issue was that the TX index is written to disk asynchronously and in our particular test, we need to wait for it to be written before we can depend on it.

There were some major formatting problems with the file.  I reformatted the entire file with .clang-format

There are only two lines changed in this file if you ignore white space.

### DIFF
--- a/src/test/assets_tests.cpp
+++ b/src/test/assets_tests.cpp
@@ -17,6 +17,7 @@
 #include <keystore.h>
 #include <spork.h>
 #include <txmempool.h>
+#include <index/txindex.h>
 
 #include <evo/specialtx.h>
 #include <evo/providertx.h>
@@ -81,7 +82,6 @@ static void FundTransaction(CMutableTransaction &tx, SimpleUTXOMap &utoxs, const
 }
 
 static void SignTransaction(const CTxMemPool &mempool, CMutableTransaction &tx, const CKey &coinbaseKey) {
-    std::this_thread::sleep_for(std::chrono::milliseconds(50)); //txindex extra time for windows
     CBasicKeyStore tempKeystore;
     tempKeystore.AddKeyPubKey(coinbaseKey, coinbaseKey.GetPubKey());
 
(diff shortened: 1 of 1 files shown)


src/test/assets_tests.cpp
```diff
@@ -5,45 +5,48 @@
...
 
-#include <script/interpreter.h>
-#include <script/standard.h>
-#include <script/sign.h>
-#include <validation.h>
-#include <chainparams.h>
 #include <base58.h>
-#include <netbase.h>
+#include <chainparams.h>
+#include <index/txindex.h>
```
CodeIsTheKey, line 10: Added this include (others were reordered based on .clang-format).

src/test/assets_tests.cpp
```diff
@@ -80,23 +83,23 @@ static void FundTransaction(CMutableTransaction &tx, SimpleUTXOMap &utoxs, const
     }
 }
 
-static void SignTransaction(const CTxMemPool &mempool, CMutableTransaction &tx, const CKey &coinbaseKey) {
-    std::this_thread::sleep_for(std::chrono::milliseconds(50)); //txindex extra time for windows
```
CodeIsTheKey, line 84: Removed workaround

src/test/assets_tests.cpp
```diff
@@ -200,900 +204,459 @@ CreateMintAssetTx(const CTxMemPool &mempool, SimpleUTXOMap &utxos, const CKey &c
...
+    BOOST_ASSERT(passetsCache->GetAssetMetaData(assetId, asset));
 
-Height()
+    BOOST_ASSERT(asset.circulatingSupply == 1000 * COIN);
 
-== nHeight + 1);
-BOOST_ASSERT(block
-->
+    // Allow TX index to catch up with the block index.
+    g_txindex->BlockUntilSyncedToCurrentChain();
```
CodeIsTheKey, line 368: This is what was missing.  Since we just created the transaction we need to reference, we need to wait for the index to be written before using it.

src/test/assets_tests.cpp
```diff
@@ -200,900 +204,459 @@ CreateMintAssetTx(const CTxMemPool &mempool, SimpleUTXOMap &utxos, const CKey &c
...
-BuildAssetTransaction(scriptPubKey);
-CTxOut out(0, scriptPubKey);
-tx2.vout.
-push_back(out);
+    BOOST_ASSERT(asset.circulatingSupply == 100 * COIN);
 
-tx2.vin.
-push_back(CTxIn(COutPoint(tx.GetHash(), 0))
-);
+    {
```
nandofw, line 445:
```suggestion
    // Allow TX index to catch up with the block index.
    g_txindex->BlockUntilSyncedToCurrentChain();
    
    {
```
here need to wait for the index too

src/test/assets_tests.cpp
```diff
@@ -200,900 +204,459 @@ CreateMintAssetTx(const CTxMemPool &mempool, SimpleUTXOMap &utxos, const CKey &c
...
-BuildAssetTransaction(scriptPubKey2);
-CTxOut out2(0, scriptPubKey2);
-tx2.vout.
-push_back(out2);
+    BOOST_ASSERT(asset.circulatingSupply == 10 * COIN);
 
-tx2.vin.
-push_back(CTxIn(COutPoint(tx.GetHash(), 0))
-);//uniqueId=0
+    {
```
nandofw, line 567:
```suggestion
    // Allow TX index to catch up with the block index.
    g_txindex->BlockUntilSyncedToCurrentChain();
    
    {
```
and here

nandofw (changes requested):

src/test/assets_tests.cpp
CodeIsTheKey: Thanks - fixed.

src/test/assets_tests.cpp
CodeIsTheKey: Thanks - fixed.

npq7721 (approved):
LGTM

nandofw (approved):

PR #309, opened 2023-08-22
auto hide the toolbar menu

nandofw:
added option to automatic hide the toolbar menu, default enabled, can be disabled under `Settings -> options -> wallet`
fixed a font issue on send pages when `Add Recipient` or `Clear All` is clicked


npq7721 (approved):
LGTM

PR #310, opened 2023-08-25
Multi thread Headers POW calculation

nandofw:
This pr implement multi thread support to calculate the POW hash reducing block headers sync time significantly.
By default will use a maximum of 8 threads or number of cores the CPU has whichever is less. Can be increased/decreased with `-powheaderthreads` flag

some benchmark from raptoreumd syncing the first 40000 block headers on a quad core i5 4460
image


CodeIsTheKey (approved):
LGTM.  Tested locally and works nicely.  I have some ideas to improve it, but this is a great improvement already, so let's get it in!

npq7721 (approved):
Awsome

PR #311, opened 2023-08-27
Fix potential crash during sync

CodeIsTheKey:
Fixed an issue where Qt code could access a core object without a lock.  If it changed during the call, it can cause the program to crash.

### DIFF
--- a/src/evo/deterministicmns.cpp
+++ b/src/evo/deterministicmns.cpp
@@ -131,8 +131,7 @@ bool CDeterministicMNList::IsMNValid(const CDeterministicMNCPtr &dmn, int height
 }
 
 bool CDeterministicMNList::IsMNValid(const CDeterministicMNCPtr &dmn) {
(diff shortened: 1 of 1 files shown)


src/evo/deterministicmns.cpp
```diff
@@ -131,8 +131,7 @@ bool CDeterministicMNList::IsMNValid(const CDeterministicMNCPtr &dmn, int height
 }
 
 bool CDeterministicMNList::IsMNValid(const CDeterministicMNCPtr &dmn) {
-    int height = ::ChainActive().Tip() == nullptr ? 0 : ::ChainActive().Tip()->nHeight;
-    return IsMNValid(dmn, height);
+    return IsMNValid(dmn, ::ChainActive().AtomicHeight());
```
CodeIsTheKey, line 134: The non-core code (i.e., Qt) is calling IsMNValid without a lock to cs_main.  Because of this, the call to .Tip() is not safe.  Switched to the AtomicHeight which was created for this very reason.

npq7721 (approved):
LGTM

PR #313, opened 2023-09-03
[Assets] unique asset

nandofw:
changes:

- removed the limit of unique asset mint quantity
- changed unique asset mint from UTXO to a range system
- increased asset transfer uniqueId from 32 bits to 64 bits
- adjusted unique assets consensus checks

### DIFF
--- a/src/consensus/tx_verify.cpp
+++ b/src/consensus/tx_verify.cpp
@@ -249,10 +249,11 @@ unsigned int GetTransactionSigOpCount(const CTransaction &tx, const CCoinsViewCa
 
     return nSigOps;
 }
-//check to see where these now
 
 inline bool checkOutput(const CTxOut &out, CValidationState &state, CAmount &nValueIn,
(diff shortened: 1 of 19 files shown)


src/consensus/tx_verify.cpp
```diff
@@ -272,24 +273,66 @@ inline bool checkOutput(const CTxOut &out, CValidationState &state, CAmount &nVa
...
+    for (auto &vin : nVinIds){
+        vin.second = combineUniqueIdPairs(vin.second);
+    }
+    
+    for (auto &vout : nVoutIds){
+        vout.second = combineUniqueIdPairs(vout.second);
+    }
+
+    //check uniqueids
+    for (auto vin : nVinIds){
```
npq7721, line 322: this loops is inefficient could cause stall node while performance this check for multiple tx with large unique ids. would be possible to hash ids in a way to prove it match?

nandofw: i think a unordered_map can be used here, let me test

npq7721 (approved):
Awesome work

PR #315, opened 2023-09-06
Update required Debian build packages

michal-zurkowski:
libxkbcommon requires `yacc` which is not available by default in the currently shown packages in Ubuntu 20.04.


npq7721 (approved):
thanks you

PR #323, opened 2023-09-21
Fixes various

nandofw:
fixed future unique asset transaction
fixed protx register_submit
asset details popup text is now selectable
added mempool checks on create asset


npq7721 (approved):
TGTM

PR #326, opened 2023-10-29
asset balance index

nandofw:
added asset balance index
added listaddressesbyasset RPC call
added listassetbalancesbyaddress RPC call
note: the amount returned on the 2 rpc call above are in sats and can be longer than 64 bits (128 bits integer)

removed code that is no longer necessary related to the max money switch to 21b, making it permanently 21b


npq7721 (approved):
LGTM

PR #331, opened 2023-12-01
Assets update

nandofw:
implemented asset update transaction
added updateasset rpc call
added update asset GUI page
adjusted consensus checks for the update asset transaction

image

### DIFF
--- a/src/evo/providertx.cpp
+++ b/src/evo/providertx.cpp
@@ -234,6 +234,10 @@ bool CheckUpdateAssetTx(const CTransaction &tx, const CBlockIndex *pindexPrev, C
         return state.DoS(100, false, REJECT_INVALID, "bad-assets-invalid-id");
     }
 
+    if (!asset.updatable) {
+        return state.DoS(100, false, REJECT_INVALID, "bad-assets-not-updateable");
+    }
+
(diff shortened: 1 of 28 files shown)


src/evo/providertx.cpp
```diff
@@ -254,6 +258,31 @@ bool CheckUpdateAssetTx(const CTransaction &tx, const CBlockIndex *pindexPrev, C
...
+
+    if (assetTx.issueFrequency <= 0 && assetTx.type != 0) {
+        return state.DoS(100, false, REJECT_INVALID, "bad-assets-issueFrequency");
+    }
+
+    if (assetTx.maxMintCount < asset.mintCount) {
+        return state.DoS(100, false, REJECT_INVALID, "bad-assets-maxMintCount");
+    }
+    //when transferring ownership don't allow any other change
+    if (assetTx.ownerAddress != asset.ownerAddress) {
```
npq7721, line 277: if this is ownership transfer, should ownerAddress be different?

nandofw: yes, as ownership is defined by the ownerAddress, they need to be different if we are transferring the ownership of the asset

npq7721 (approved):
LGTM

PR #338, opened 2024-02-02
Update build-unix.md

bertux:
for Ubuntu packages install
As draft because I'm testing it on my laptop.

### DIFF
--- a/doc/build-unix.md
+++ b/doc/build-unix.md
@@ -12,7 +12,7 @@ Run the following commands to install required packages:
 
 ##### Debian/Ubuntu:
 ```bash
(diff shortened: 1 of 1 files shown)


michal-zurkowski:
What version of Ubuntu are you using?
Just tested it on fresh 22.04 & 20.04 (2 currently supported versions) Desktop and Server. Followed steps for stable (current mainnet version) as well as steps for develop (current testnet) and both were built just fine using the provided packages.
The latest addition of `bison` was added to support the requirement of `yacc`
I'm also not sure why you need to add `cmake` as it is compiled and used with all the other dependencies and systemwide `cmake` will never be used.

npq7721 (approved):
LGTM

Issue #336, opened 2024-01-30
Fix network totals display (QT | Testnet)


Raptor3um:
Testnet build version: Latest from actions
OS: Win 11

Currently in QT > Window > Network Traffic: Network totals display is broken.
![testnet-network totals](https://github.com/Raptor3um/raptoreum/assets/47279850/f6641694-30d2-4926-b3b0-b285f4fcedb4)

How it looks on mainnet:
![mainnet-network totals](https://github.com/Raptor3um/raptoreum/assets/47279850/56802600-ff29-4f98-af88-747ee05039a9)

nandofw:
fixed on pr #339

PR #339, opened 2024-02-02
fixes

nandofw:
fixed display issue on command-line options page
reimplemented commit #288 
fixed network totals display issue #336 
fixed asset balance not updating properly
fixed dust issue on transfer asset output
fixed a mempool issue when create sub asset

### DIFF
--- a/src/addressindex.h
+++ b/src/addressindex.h
@@ -68,7 +68,7 @@ struct CMempoolAddressDeltaKey {
     CMempoolAddressDeltaKey(int addressType, uint160 addressHash) {
         type = addressType;
         addressBytes = addressHash;
-        asset = "RTM";
+        asset.clear();
         txhash.SetNull();
         index = 0;
         spending = 0;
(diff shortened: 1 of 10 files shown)


src/addressindex.h
```diff
@@ -68,7 +68,7 @@ struct CMempoolAddressDeltaKey {
     CMempoolAddressDeltaKey(int addressType, uint160 addressHash) {
         type = addressType;
         addressBytes = addressHash;
-        asset = "RTM";
```
npq7721, line 71: was this added on other PR to fix getmempooladdresses calls?

nandofw: yes

npq7721: then it getting removed in this one and replace with asset,clear(). is it setting it somewhere else?

nandofw: on this constructor the name should be empty, on the last pr i set it wrong and i only discover that it was wrong too late

npq7721 (approved):
LGTM

PR #340, opened 2024-02-02
move event connections and hide unneeded fields

### LINKED ISSUE #333: Remove custom change address from CoinJoin tab
CoinJoin spends inputs completely, any leftover is added to fee to ensure all is spent, because of this there is a check / stop in place to prevent accidentally sending with a overly large fee.

The "Custom Change Address" field and select box should be removed from the CoinJoin tab. 

This is present on both current mainnet release and current testnet release.

remove change address option

bertux:
I've verified on my build to fix #333


npq7721 (approved):
LGTM

PR #341, opened 2024-02-08
Initial Round Voting Implementation

CodeIsTheKey:
Round Voting initial implementation.  Miner voting tested and working.  Node voting implementation in progress (do not enable).


src/update/update.h
```diff
@@ -0,0 +1,428 @@
...
+         (
+            (bit < 0 || bit > 28)                                         ||
+            (roundSize < 1)                                               ||
+            (startHeight < 0)                                             ||
+            (votingPeriod < 1)                                            ||
+            (votingMaxRounds < 1)                                         ||
+            (graceRounds < 0)                                             ||
+            (heightActivated < -1)                                        ||
+            (startHeight % roundSize != 0)                                ||
+            (votingPeriod > votingMaxRounds)
```
nandofw, line 110: is this line correct?
note: it fail on launch as votingPeriod is 4031 and votingMaxRounds is 3

src/llmq/quorums_commitment.cpp
```diff
@@ -86,8 +87,14 @@ namespace llmq {
 
         // sigs are only checked when the block is processed
         if (checkSigs && IsBlsSigCheckEnabled(pQuorumBaseBlockIndex->GetBlockTime())) {
-            uint256 commitmentHash = CLLMQUtils::BuildCommitmentHash(llmq_params.type, quorumHash, validMembers,
-                                                                     quorumPublicKey, quorumVvecHash);
+
+            uint256 commitmentHash;
+            if (UpdateManager::Instance().IsActive(EUpdate::ROUND_VOTING, pQuorumBaseBlockIndex)) {
+                commitmentHash = CLLMQUtils::BuildCommitmentHash(llmqType, quorumHash, validMembers, quorumUpdateVotes, quorumPublicKey, quorumVvecHash);
```
nandofw:
```suggestion
                commitmentHash = CLLMQUtils::BuildCommitmentHash(llmq_params.type, quorumHash, validMembers, quorumUpdateVotes, quorumPublicKey, quorumVvecHash);
```

nandofw:
```suggestion
                commitmentHash = CLLMQUtils::BuildCommitmentHash(llmq_params.type, quorumHash, validMembers, quorumUpdateVotes, quorumPublicKey, quorumVvecHash);
```

src/update/update.cpp
```diff
@@ -0,0 +1,517 @@
...
+   minerUpdateVoting(&minerRoundVoting),
+   nodeUpdateVoting(&nodeRoundVoting)
+{
+}
+
+UpdateManager::~UpdateManager() {};
+
+bool UpdateManager::Add(Update update)
+{
+   // TODO: Check for existence first
```
nandofw:
```suggestion
  //Check for existence first and erase if exist
   auto it = updates.find(update.UpdateId());
   if (it != updates.end())
      updates.erase(it);
```
during start up 4 chains are initialized and the last chain initialized is the active chain, we only need update settings for the active chain

nandofw (changes requested):

src/update/update.h
CodeIsTheKey: Good catch - the mainnet votingMaxRounds should be > 4031 (I'll update to 3*4031 to safely cover the range)

src/llmq/quorums_commitment.cpp
CodeIsTheKey: Will update - logically should be the same, but better to keep consistent.

src/update/update.cpp
CodeIsTheKey: Will fix - thanks.

CodeIsTheKey:
Updated from PR comments.  Removed raptoreum-tx from build.yaml, minor typo fix.

nandofw (approved):

CodeIsTheKey:
Fixed assetindex problem when loading block index.

PR #366, opened 2024-03-17
[Asset] clean up

nandofw:
removed v1 asset implementation to align mainnet/testnet code base
moved IsAssetsActive to the update manager to use soft-fork to activate assets on the upcoming testnet


npq7721 (approved):
LGTM

PR #368, opened 2024-04-04
Ft/asset rip01 testnet

npq7721:


nandofw (approved):
LGTM

CodeIsTheKey:
Looks good to me.  Voting is enabled in approx. 2 days.  Round votes are roughly daily and we take the last 7 round votes to form an update vote.  Let 'er rip!

PR #375, opened 2024-05-17
Ft/uppercase

npq7721:
bump version from 2.0.0.99 to 2.0.1.99
asset name field during asset creation is now case insensitive


nandofw (approved):
LGTM

PR #389, opened 2024-08-12
Ft/wallet design

drmancini:
Introduction of a new dark/light visual theme fro the Qt wallet

1. Dark theme set as default because it's prettier
2. An updated Light theme is probably coming later
3. CoinJoin is disabled by default as I believe most users don't even know what this is
4. Toolbar hiding is disabled by default because hidden toolbar for new users is a source of pain. And pain leads to suffering.
5. Icon and Image cleanup

New design created by Zlata so give her credit too ;-)


npq7721 (approved):
Great works

PR #394, opened 2024-08-17
Prevent escape from "closing" a dialog, introducing a blank window.

CodeIsTheKey:
In the Create, Update, and View Assets pages, pressing escape while in a text box would cause the dialog to disappear, leaving a blank window for the user.


npq7721 (approved):
LGTM

PR #398, opened 2024-08-21
Fix compile with -disable-wallet

ericM80:
#define out the wallet-specific functions from rpcassets.cpp:
createasset updateasset mintasset sendasset

fix bug in rpcevo.cpp, if statement starting with else when not ENABLE_WALLET


npq7721 (approved):
LGTM

PR #418, opened 2025-01-21
Fixed Root asset handling with all languages.

CodeIsTheKey:
Root asset logic in the Qt wallet was comparing the text of the combo box instead of the index.  Fixed.


npq7721 (approved):
LGTM

PR #421, opened 2025-06-21
Bug/pruning with voting active

CodeIsTheKey:
The chain cannot be pruned beyond the start of the oldest active vote.  I marked the last vote as complete to resolve this.


npq7721 (approved):
LGTM

PR #424, opened 2025-08-08
rpc: Add 'mine' flag to listassets command (fixes #422)

### LINKED ISSUE #422: add mine flag in listassets
To be able to list only assets belong to existing wallet in core wallet we need add mine flag into listassets command.

mine flag default is false, if true will only return assets that belong to current wallet from raptoreum core wallet.

halfaipg:
- Add optional 'mine' parameter (default: false) to listassets RPC
- When mine=true, only return assets owned by the current wallet
- Implement tolerant parsing for mine parameter (bool/number/string)
- Update help examples to show mine flag usage
- Add release notes documentation


npq7721 (approved):
LGTM

PR #429, opened 2025-11-29
Fixed potential race condition and crash when cleaning up old quorum contributions.

CodeIsTheKey:
This fixes a crash when cleaning up old quorum contributions and avoid potential race conditions during maintenance.


npq7721 (approved):
LGTM

PR #432, opened 2025-12-04
Fix typo in compiler flag (-Woverloded-virtual → -Woverloaded-virtual)

Germardies:
This PR fixes a build error caused by a misspelled compiler warning flag
(-Woverloded-virtual → -Woverloaded-virtual) in configure.ac.

The incorrect flag prevents compilation on modern GCC versions:

g++: error: unrecognized command-line option '-Woverloded-virtual'

This change corrects the flag and restores compatibility. Build tested on Ubuntu 22.04.


JSanchezFDZ (approved):
Verified the fix. `AX_CHECK_COMPILE_FLAG` probes `-Woverloaded-virtual`, but the
success branch was appending the misspelled `-Woverloded-virtual` to
`WARN_CXXFLAGS`, which GCC then rejects with `unrecognized command-line option`,
breaking the build on modern toolchains. The one-character correction makes the
appended flag match the probed one. Correct and complete. LGTM.

PR #437, opened 2026-02-27
Entering optional fee addresses for assets.

Germardies:
Optional fee addresses have been added to all asset instructions in the Raptoreum CLI.
An asset change address has been added specifically for sendasset.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -161,3 +161,5 @@ dist/
 osx_volname
 dist/
 src/test/runtest.sh
+configure~
+src/dashbls/configure~
--- a/configure.ac
+++ b/configure.ac
@@ -411,7 +411,7 @@ if test "$CXXFLAGS_overridden" = "no"; then
   AX_CHECK_COMPILE_FLAG([-Wduplicated-branches],       [WARN_CXXFLAGS="$WARN_CXXFLAGS -Wduplicated-branches"],       [], [$CXXFLAG_WERROR])
   AX_CHECK_COMPILE_FLAG([-Wduplicated-cond],           [WARN_CXXFLAGS="$WARN_CXXFLAGS -Wduplicated-cond"],           [], [$CXXFLAG_WERROR])
   AX_CHECK_COMPILE_FLAG([-Wlogical-op],                [WARN_CXXFLAGS="$WARN_CXXFLAGS -Wlogical-op"],                [], [$CXXFLAG_WERROR])
-  AX_CHECK_COMPILE_FLAG([-Woverloaded-virtual],        [WARN_CXXFLAGS="$WARN_CXXFLAGS -Woverloded-virtual"],         [], [$CXXFLAG_WERROR])
+  AX_CHECK_COMPILE_FLAG([-Woverloaded-virtual],        [WARN_CXXFLAGS="$WARN_CXXFLAGS -Woverloaded-virtual"],         [], [$CXXFLAG_WERROR])
   dnl -Wsuggest-override is broken with GCC before 9.2
   dnl https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78010
   AX_CHECK_COMPILE_FLAG([-Wsuggest-override], [WARN_CXXFLAGS="$WARN_CXXFLAGS -Wsuggest-override"],                               [], [$CXXFLAG_WERROR],
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -729,6 +729,7 @@ libraptoreum_util_a_SOURCES = \
   threadinterrupt.cpp \
   util/bip32.cpp \
   util/system.cpp \
+  bootstrap/bootstrapmanager.cpp \
   util/asmap.cpp \
   util/moneystr.cpp \
   util/spanparsing.cpp \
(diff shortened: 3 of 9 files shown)


JSanchezFDZ (commented):
Reviewed the optional asset fee-address feature.

1. **Scope / duplicate.** This PR's head (eb4a1ea0) is the same commit as #438, so
   it also contains the entire auto-bootstrap downloader plus `lockedpool.cpp`,
   `bip32.h`, and `.gitignore` changes. The actual fee-address change is only part
   of `rpc/rpcassets.cpp`. Please rebase so this PR contains just the `feeAddress`
   feature and close/separate #438 — otherwise the (security-sensitive) bootstrap
   code blocks an otherwise small asset feature. The `bip32.h <cstdint>` and
   `lockedpool.cpp <stdexcept>` additions are legitimate build fixes but unrelated;
   fine to keep, but please call them out or split them.

2. **Feature itself.** The approach across createasset/sendasset/mintasset
   (`DecodeDestination` + `IsValidDestination`, `coinControl.destChange = feeDest`,
   manual selection of `feeDest` UTXOs, `fAllowOtherInputs = false`) reads
   reasonably. Two things to confirm:
   - With `fAllowOtherInputs = false` and only `feeDest` UTXOs selected, please
     verify the asset funding inputs/outputs still resolve for all three RPCs — a
     `feeAddress` that holds RTM but where the asset/owner outputs would need other
     inputs could fail in a way users won't expect.
   - The "no spendable UTXOs on feeAddress" path throws
     `RPC_WALLET_INSUFFICIENT_FUNDS`, which is good; a functional test covering
     that path and "change returns to feeAddress" would make this much easier to
     merge with confidence.

JSanchezFDZ (changes requested):
Reviewed `eb4a1ea06ab9e913a3282fc91342c836ba25fced` against `900794ea94ff11023667765173295b459d32732c`: complete nine-file diff, with detailed tracing of asset RPCs, coin control, RTM/asset selection, transaction creation, and named-argument dispatch. **Requesting changes for two wallet/RPC issues and the shared bootstrap blockers.**

**1. P2 — The fee address forces every matching UTXO into the transaction.**

The selection loops in [rpcassets.cpp:275–287](https://github.com/Raptor3um/raptoreum/blob/eb4a1ea06ab9e913a3282fc91342c836ba25fced/src/rpc/rpcassets.cpp#L275), and the corresponding `updateasset`, `mintasset`, and `sendasset` paths, call `coinControl.Select` for every UTXO at the address. They set `fAllowOtherInputs=false` but leave `fRequireAllInputs=true`. In `SelectCoins`, that combination consumes every selected spendable input instead of stopping when the required amount is reached.

A fragmented fee address can therefore make an otherwise fundable transaction exceed `MAX_STANDARD_TX_SIZE`, or incur unnecessary consolidation fees. For example, 700 ordinary compressed-key P2PKH inputs already exceed 100,000 bytes, even when one would pay the fee. I traced the actual selection and size-check branches and checked a small equivalent model; I did not build/send a real transaction. Set `fRequireAllInputs=false` in these four paths, preserving the address restriction. `FundSpecialTx` already uses that pattern.

**2. P2 — Register `fee_address` as a named argument.**

The new positional parameters at [rpcassets.cpp:637–638](https://github.com/Raptor3um/raptoreum/blob/eb4a1ea06ab9e913a3282fc91342c836ba25fced/src/rpc/rpcassets.cpp#L637) and [799–800](https://github.com/Raptor3um/raptoreum/blob/eb4a1ea06ab9e913a3282fc91342c836ba25fced/src/rpc/rpcassets.cpp#L799) are missing from the [CRPCCommand argument lists](https://github.com/Raptor3um/raptoreum/blob/eb4a1ea06ab9e913a3282fc91342c836ba25fced/src/rpc/rpcassets.cpp#L1427). `mintasset` and `sendasset` therefore reject JSON-object parameters containing `fee_address` with `Unknown named parameter fee_address`, before entering their handlers. I traced `ExecuteCommand`/`transformNamedArguments` and verified the rejection using names extracted from the actual table. Add the name to both registrations without removing existing names.

For `sendasset`, also cover omitted intermediate change addresses: named-argument conversion fills those positions with `null`, which their existing `get_str()` calls do not accept. That hole-handling limitation predates this PR, but needs accounting for when exposing the new trailing named parameter. `createasset`/`updateasset` use the already registered `asset` object and do not have this registration defect.

**Shared bootstrap blockers**

#437 and #438 have identical HEADs and diffs. The [detailed bootstrap review on #438](https://github.com/Raptor3um/raptoreum/pull/438) also applies here: unconstrained extraction, trusting remotely supplied PoW-cache values, bypassing the configured proxy, treating pruned/external-block directories as empty, incomplete Qt/dependency integration, wrong-network imports, partial-install/error handling, missing cancellation, and source-distribution omissions. I independently checked the principal source traces. These cannot be excluded from this PR solely because its title describes fee addresses. Separating the two features would make their scope and integration clearer.

**Validation and limits**

- Read all four handlers, coin-control defaults, both selection paths, signing/commit, and dispatcher callers. The model confirms mandatory selection of 700 inputs versus one sufficient input when the flag is disabled, and both named-argument rejections.
- Checked that RTM inputs from other addresses remain excluded. Asset inputs have a separate selection set; `fAllowOtherInputs=false` does not by itself prevent selecting the asset. Locked, watch-only, and unavailable future inputs retain their spendability checks.
- Verified identical #437/#438 trees and the merge base. No node, transaction broadcast, complete C++ build, or bootstrap download was executed; the selection proof combines actual source tracing with a model. No check runs were available for this SHA in the query.

Add tests for fragmented fee addresses in all four RPCs, positional/named forms with omitted/explicit change addresses, assets held separately from the fee address, and insufficient/locked/watch-only/future funds. Document that this address funds burn fees as well as network fees, and distinguish RTM change from asset change. I found no additional measured performance issue warranting a new index or abstraction.

Issue #151, opened 2022-04-01
Add libcurl to Depends


CodeIsTheKey:
We have a few needs for libcurl in order to pull files directly with the core wallet.  Adding libcurl to depends so that the library is available to use for Linux, Windows, and MacOS would be very helpful.

We should add the most recent, compatible version that is available.

PR #438, opened 2026-03-01
Proposal: Add automatic bootstrap download on first run for raptoreumd

Germardies:
## Automatic Bootstrap Download on First Run for raptoreumd (no QT)

This PR adds automatic blockchain bootstrap downloading when Raptoreum Core is started for the first time with an empty data directory.

### What this does
- Detects first run by checking if `blocks/blk00000.dat` is missing
- Automatically downloads `bootstrap.zip` from bootstrap.raptoreum.com
- Extracts the archive into the data directory
- Downloads `powcache.dat` for faster initial 

This is purely a suggestion for improvement!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -161,3 +161,5 @@ dist/
 osx_volname
 dist/
 src/test/runtest.sh
+configure~
+src/dashbls/configure~
--- a/configure.ac
+++ b/configure.ac
@@ -411,7 +411,7 @@ if test "$CXXFLAGS_overridden" = "no"; then
   AX_CHECK_COMPILE_FLAG([-Wduplicated-branches],       [WARN_CXXFLAGS="$WARN_CXXFLAGS -Wduplicated-branches"],       [], [$CXXFLAG_WERROR])
   AX_CHECK_COMPILE_FLAG([-Wduplicated-cond],           [WARN_CXXFLAGS="$WARN_CXXFLAGS -Wduplicated-cond"],           [], [$CXXFLAG_WERROR])
   AX_CHECK_COMPILE_FLAG([-Wlogical-op],                [WARN_CXXFLAGS="$WARN_CXXFLAGS -Wlogical-op"],                [], [$CXXFLAG_WERROR])
-  AX_CHECK_COMPILE_FLAG([-Woverloaded-virtual],        [WARN_CXXFLAGS="$WARN_CXXFLAGS -Woverloded-virtual"],         [], [$CXXFLAG_WERROR])
+  AX_CHECK_COMPILE_FLAG([-Woverloaded-virtual],        [WARN_CXXFLAGS="$WARN_CXXFLAGS -Woverloaded-virtual"],         [], [$CXXFLAG_WERROR])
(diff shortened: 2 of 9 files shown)


JSanchezFDZ (commented):
Reviewed the auto-bootstrap downloader. The intent is useful, but there are a few
blocking concerns, mostly around security:

1. **Duplicate / mixed scope.** This PR's head (`feature/auto-bootstrap`,
   eb4a1ea0) is the exact same commit as #437, and it bundles two unrelated
   features — the asset `feeAddress` work in `rpc/rpcassets.cpp` and the bootstrap
   downloader — plus incidental changes (`lockedpool.cpp`, `bip32.h`, `.gitignore`).
   Please split into one focused PR per feature and close the duplicate; as-is
   neither change can be reviewed or merged cleanly.

2. **No integrity/authenticity check (security).** `DownloadFile()` fetches
   `bootstrap.zip` / `powcache.dat` over HTTPS but nothing verifies the payload
   (no SHA-256, no signature). A compromised or MITM'd bootstrap host could feed a
   crafted snapshot. Please verify a published, ideally signed, SHA-256 of the
   archive before extracting and abort on mismatch.

3. **Path traversal / zip-slip (security).** `ExtractArchive()` uses
   `archive_write_disk_new()` without `archive_write_disk_set_options(...)`. Set
   `ARCHIVE_EXTRACT_SECURE_NODOTDOT | ARCHIVE_EXTRACT_SECURE_SYMLINKS |
   ARCHIVE_EXTRACT_SECURE_NOABSOLUTEPATHS`; otherwise an archive entry with `../`,
   an absolute path, or a symlink can write outside the data directory. The manual
   `destDir + "/" + archive_entry_pathname()` concatenation does not prevent this.

4. **Error handling.** Return values of `archive_write_header()` /
   `archive_write_data_block()` are ignored, and `ExtractArchive()` returns `true`
   even when the read loop aborts on error, so it can report "Extraction complete"
   after a partial/failed extract. `DownloadFile()` also doesn't set
   `CURLOPT_FAILONERROR`, so an HTTP 4xx/5xx error page is written into
   `bootstrap.zip` and only fails later at extraction. Consider `CURLOPT_FAILONERROR`
   and explicit `CURLOPT_SSL_VERIFYPEER`/`CURLOPT_SSL_VERIFYHOST`.

5. **New daemon dependencies.** This adds `-lcurl -larchive` to `raptoreumd`,
   which is a meaningful new attack/build surface and complicates depends and
   reproducible builds (related to #151). Worth a maintainer decision before
   pulling libcurl into the daemon.

6. **Opt-in/opt-out.** It triggers automatically on first run (detecting a missing
   `blocks/blk00000.dat`). Please make it opt-in or at least disableable via a flag,
   and note that `-reindex`, pruned, or custom datadir layouts may misdetect a
   "first run".

JSanchezFDZ (changes requested):
Review status changed to **Request changes**. The findings, evidence, and validation limits are consolidated in the [original review](https://github.com/Raptor3um/raptoreum/pull/438#pullrequestreview-4545337105). Please use that review for the technical discussion.

Issue #393, opened 2024-08-17
?(help) is no help at all


Raptor3um:
**Chain:** Testnet
**Branch:** Develop
**Build:** v2.0.03.99

All windows which have a ? for "help" have a bug where clicking on the ? has an unexpected result (see video).

**Note:** This appears to be a Windows OS specific issue.

https://github.com/user-attachments/assets/9f6fb899-2300-47b8-ace0-975a3f67cace

There is a bounty of 25k RTM on this issue.

drmancini:
fixed with proposed PR #402

HumbleDeer:
This issue is currently still present in the wallet, but the "red crossed circle" pointer can be exited by clicking any clickable user interface element (buttons etc)

Bigpiggy01:
Can confirm this is still present on Windows. So still open!

Clicking on it now makes it look like a forbidden/invisible windows popsup. Cursor turns into a red forbidden sign.

PR #441, opened 2026-04-17
Update utilitydialog.cpp

Balin0x:
Should remove the [?] which is in issue : https://github.com/Raptor3um/raptoreum/issues/393 
image


CodeIsTheKey:
Update looks good.  There are other dialogs that have the same issue, but this fixes this one.

CodeIsTheKey (approved):

CodeIsTheKey:
We should merge this into develop first.

JSanchezFDZ (approved):
Standard fix for the stray Windows context-help "[?]" title-bar button (issue
#393): clearing `Qt::WindowContextHelpButtonHint` is the conventional approach
and is correctly scoped to Windows via `Q_OS_WIN`, so other platforms are
unaffected. LGTM.

PR #443, opened 2026-05-12
EVM integration — 32 commits across Phases 0–5 (draft for review)

JSanchezFDZ:
## Summary

Native EVM integration. 32 commits across 5 phases: account
abstraction layer, full evmone-backed execution, ConnectBlock /
DisconnectTip wiring, reorg journal, EIP-1559 fee accounting,
parallel pre-flight, complete `eth_*`/`net_*`/`web3_*` JSON-RPC
namespace (22 methods + default 8545 listener), four RTM-native
precompiles (ChainLocks, Masternode Registry, Smart Asset ERC-20,
LLMQ Oracle), and C++ EIP-1559 signing primitives byte-for-byte
identical to Python `eth-account` on canonical test vectors.

**Marked DRAFT — this is for architecture review, not merge.**
Phases 6 (testnet + paid audits) and 7 (mainnet activation) live
in the plan after the review feedback lands.

## Why review now

- Validate the architecture before going deeper (Phase 6 paid
  audits are expensive).
- Resolve 10 open questions still owed by the core team
  (`Q-A1`..`Q-A10` in [`docs/evm/QUESTIONS.md`](docs/evm/QUESTIONS.md)) — these unblock
  Phase 5.4 (wallet HD), Phase 6 (testnet), Phase 7 (mainnet).
- Get directional feedback on the four precompiles, the
  consensus carve-outs in `tx_check`/`validation`/`chainparams`,
  the receipt persistence, and the D2 header-field deferral.

## Documentation entry point

Start with [`docs/evm/README.md`](docs/evm/README.md). It indexes:

- **[`STATUS.md`](docs/evm/STATUS.md)** — phase-by-phase map with commit refs and test surface
- **[`RPC.md`](docs/evm/RPC.md)** — all 25 RPC methods with examples
- **[`PRECOMPILES.md`](docs/evm/PRECOMPILES.md)** — Solidity interfaces, selectors, addresses, gas costs
- **[`FUP.md`](docs/evm/FUP.md)** — follow-up register (mainnet-blocking vs. before-testnet vs. post-launch)
- **[`QUESTIONS.md`](docs/evm/QUESTIONS.md)** — Q-A1..Q-A10 still open with the core team
- **[`BUILD.md`](docs/evm/BUILD.md)** — build / run / test guide + full file inventory
- **[`PLAN.md`](docs/evm/PLAN.md)** + **[`PROPOSAL-FOR-CORE-TEAM.md`](docs/evm/PROPOSAL-FOR-CORE-TEAM.md)** — historical reference (the original plan and the formal acceptance with D1–D7 decisions)

## What ships in this PR

| Phase | Status | Highlights |
|---|---|---|
| **0** evmone spike | ✅ | `evm_executeReadOnly` smoke RPC, smoke test, depends/ wiring |
| **1** AAL | ✅ | TX types 11/12/13, opcodes 0xbd/0xbe/0xbf, activation gate via UpdateManager |
| **2** Full EVM execution | ✅ | `CEvmStateDB`/`CEvmStateCache` + evmc::Host with nested CALL/CREATE + ApplyEvm{Call,Deploy,Spend}Tx + EIP-1559 ProcessEvm*Tx + ConnectBlock/DisconnectTip wiring + reorg journal + parallel pre-flight |
| **3** JSON-RPC namespace | ✅ | 22 `eth_*`/`net_*`/`web3_*` methods, `eth_sendRawTransaction` with EIP-1559+legacy decode + secp256k1 sender recovery, receipt persistence with eth↔rtm cross-index, `eth_getLogs` with filter, default 8545 listener |
| **4** RTM-native precompiles | ✅ | ChainLocks (`0x…0a03`), Masternode Registry (`0x…0a04`), Smart Asset ERC-20 (`0xA55E70…hash160`), LLMQ Oracle (`0x…0a02`) |
| **5** Signing primitives | ✅ partial | C++ EIP-1559 signing byte-identical to eth-account; HD wallet integration + Qt UI deferred (need `Q-A4`) |

**Test surface**: 11 EVM unit-test suites, all green every commit.
Non-EVM tests unaffected. End-to-end live-validated against a
regtest daemon for every RPC method. Per-suite case/assertion
counts in [`STATUS.md`](docs/evm/STATUS.md).

## Consensus-affecting changes (the diff that matters)

These are intentionally narrow carve-outs for the new EVM tx types;
they are *necessary* (an EVM tx has no UTXO inputs/outputs by
design) and are gated by tx-type checks so non-EVM transactions
behave identically.

- `src/consensus/tx_check.cpp` — `allowEmptyTxInOut` extended to
  `TRANSACTION_EVM_DEPLOY` / `_CALL` / `_SPEND` (matches the
  existing pattern for `TRANSACTION_QUORUM_COMMITMENT`).
- `src/validation.cpp` — `bad-txns-type` allow-list extended; relay-
  fee + mempool-min-fee skipped for EVM tx types (fee lives on the
  EVM side); ConnectBlock dispatches EVM-typed txs through
  `ProcessEvmTransactionsInBlock` behind `Updates().IsEvmActive()`;
  ConnectTip generates + persists receipts; DisconnectTip applies
  + erases the reorg journal.
- `src/chainparams.cpp` — `Update(EUpdate::EVM, …,
  heightActivated=0)` registered for REGTEST. Mainnet/testnet
  activation heights are coordinated hard-fork decisions (see
  `Q-A5`).

The CBlockHeader is **NOT** modified in this PR. The D2 header
fields (`stateRoot`, `receiptsRoot`, `transactionsRoot`,
`chainLocksCommit`) land alongside the mainnet activation hard fork
— tracked as `FUP-2.1` and `Q-A6`.

## What's deferred (and why)

- **`stateRoot` / `receiptsRoot` / `transactionsRoot` /
  `chainLocksCommit` header fields (D2).** Coordinated hard-fork;
  needs `Q-A6` resolution. While deferred, EIP-1559 `baseFee`
  stays at 0 and coinbase-tip-output verification is unenforced
  (`FUP-2.1` → `FUP-2.3`). All mainnet-blocking.
- **Wallet-keystore HD integration.** Awaits `Q-A4` (`m/44'/60'`
  Ethereum-standard vs `m/44'/10226'/0'/1/i` RTM-subtree). Signing
  primitives are ready; this is just the "select-key-from-wallet"
  wiring.
- **Qt UI for deploy/call dialogs.** Separate session — Qt + MOC.
- **Smart Asset ↔ EVM bidirectional balance mirror** (D4-revised
  "T-mirror" gate). Highest-risk integration piece still ahead.
  The ERC-20 precompile's `transfer`/`approve` surface stays
  unimplemented until the mirror lands. Read surface
  (`name`/`symbol`/`decimals`/`totalSupply`) is live.
- **Async `requestSignature`** on the LLMQ Oracle. Needs the gas-
  escrow + timeout pattern (`Q-A8`). Sync `verifySignature` is
  live.

Full register in [`docs/evm/FUP.md`](docs/evm/FUP.md).

## Open questions for the core team

10 questions in [`docs/evm/QUESTIONS.md`](docs/evm/QUESTIONS.md). The four that unblock immediate
next steps:

- **`Q-A4`** — HD path for EVM keys (`m/44'/60'` vs RTM coin_type).
  Unblocks Phase 5.4.
- **`Q-A5`** — Activation height per network (testnet + mainnet).
  Unblocks Phase 6.
- **`Q-A6`** — D2 header field layout + activation mechanism.
  Unblocks mainnet (`FUP-2.1`..`FUP-2.3`).
- **`Q-A7`** — Mempool gas-price floor for non-UTXO-fee EVM txs.
  DoS-defence for mainnet.

The other six are either no-rush (`Q-A1` clarification, `Q-A2/3`
asset-fee semantics, `Q-A9` MN IPv6, `Q-A10` bug-bounty multisig)
or design-tier (`Q-A8` async LLMQ).

## How to validate

1. Read [`docs/evm/STATUS.md`](docs/evm/STATUS.md) for the high-level shape.
2. Check out the branch + build per [`docs/evm/BUILD.md`](docs/evm/BUILD.md) (Docker
   instructions included; full build is ~10 minutes from clean).
3. Run all 11 EVM unit-test suites — all should pass.
4. Boot a regtest daemon (`EVM` is force-active at height 0); poke
   at the RPC namespace per [`docs/evm/RPC.md`](docs/evm/RPC.md) examples.
5. Optionally compare `evm_signTransaction` output against Python
   `eth-account` for the canonical `0x4646…4646` key — bytes
   should match exactly.

## Test plan

- [x] All 11 EVM unit-test suites green
- [x] Non-EVM unit-test suites unaffected (spot-checked: assets,
      bls, bip32, transaction)
- [x] End-to-end live regtest validation for every RPC method
- [x] EIP-1559 signing byte-parity with `eth-account` on canonical
      test vectors
- [ ] **Pending — Phase 6**: 10000+ Ethereum tests suite vectors
      (T1)
- [ ] **Pending — Phase 6**: state-divergence harness across two
      regtest nodes (T2) — requires `FUP-2.1` header fields first
- [ ] **Pending — Phase 6**: reorg fuzzing property-based tests
      (T3)
- [ ] **Pending — Phase 6**: T-mirror harness once `FUP-4.2` lands
- [ ] **Pending — Phase 6**: paid auditorías (Trail of Bits,
      ChainSecurity, OpenZeppelin)

## Branch layout

The PR is 32 commits, organised per-phase with substantive commit
bodies (the implementation notes for each phase live in the commit
messages — not just in the docs). Phases / sub-phases are visible
in `git log --oneline`:

```
8926f8785 docs(evm): comprehensive reference + status documentation
fdd0846b6 feat(evm): Phase 5 — server-side EIP-1559 signing primitives + 3 RPC methods
26a6933b9 feat(evm): Phase 4.1 + 4.2 + 4.4 — Masternodes / Smart Assets ERC-20 / LLMQ Oracle precompiles
0454a95a9 feat(evm): Phase 4.0 + 4.3 — precompile framework + ChainLocks precompile
45a3f445e feat(evm): Phase 3.7 — second HTTP listener on the Ethereum default port
2f6cd4982 feat(evm): Phase 3.6d — eth_getLogs (filter by block range / address / topics)
57d3bbec4 feat(evm): Phase 3.6 — receipt persistence + eth_getTransactionReceipt
a6b635f4a feat(evm): Phase 3.5 — eth_sendRawTransaction (RLP + EIP-1559 + sender recovery)
06b244a6e feat(evm): Phase 3.4 — Ethereum/Web3 metadata methods
285df40a5 feat(evm): Phase 3.3 — eth_getBlockByNumber / eth_getBlockByHash + counts
db5ded10d feat(evm): Phase 3.2 — eth_call + eth_estimateGas
24b3f1e67 feat(evm): Phase 3.1 — eth_* JSON-RPC namespace (read-only subset)
26328f77e feat(evm): Phase 2.5 — worker pool (D1, MVP)
c33d4f2d7 feat(evm): Phase 2.6-2 — wire reorg journaling into Connect/Disconnect
9fcac2814 feat(evm): Phase 2.6-1 — reorg journal data structures
90fb43f08 feat(evm): Phase 2.4e-2 — ConnectBlock wiring + pevmstatedb lifecycle
53891ae97 feat(evm): Phase 2.4e-1 — block-level EVM tx iterator
fb8dd3ec4 feat(evm): Phase 2.4 — ProcessEvm*Tx (EIP-1559 fee accounting)
4a3e1d384 feat(evm): Phase 2.3d — snapshot/revert + nested CALL/CREATE
1eb006846 feat(evm): Phase 2.3c — ApplyEvmSpendTx (EVM → UTXO move)
24b2a93aa feat(evm): Phase 2.3b — ApplyEvmDeployTx + real Keccak-256 + CREATE address
ae8e6a292 feat(evm): Phase 2.3a — ApplyEvmCallTx (single-call execution path)
dad0d3069 feat(evm): Phase 2.2 — evmc::Host implementation over CEvmStateCache
a48d22313 feat(evm): Phase 2.1 — state storage primitives (account, db, cache)
b853e7e8f feat(evm): Phase 1.5 — validation surface completion + ops helper
a4f9b17b7 feat(evm): Phase 1.4 — activation gate wired to UpdateManager
0386419b4 docs(evm): core team acceptance + D4 revised + D7 added
f34bcdf41 test(evm): add evmtx serialization tests + SCRIPT_ENABLE_EVM_OPCODES gating
f434a9470 build: fix NATPMP_LIBS detection in configure.ac
80c7e262c feat(evm): Phase 1.1 scaffolding — tx types, opcodes, validation stubs
51d38a437 build: add .gitattributes to enforce LF on build-critical files
c3909e50d feat(evm): Phase 0 spike — validate evmone integration
```

If you'd prefer this split across multiple PRs (e.g., one per
phase), happy to refactor. The single-PR shape keeps the
architectural narrative coherent, but it's a tradeoff against
review-chunk size.


JSanchezFDZ:
### Note on the test suite — pre-existing `ECC_Start` failure on full-suite runs

While preparing this PR I ran the full `test_raptoreum` binary and
hit a fixture-level abort that is **not** introduced by this work
but worth surfacing.

#### What happens

After the first one or two suites pass, every subsequent test
case aborts in its fixture constructor with:

```
test_raptoreum: key.cpp:340: void ECC_Start():
    Assertion `secp256k1_context_sign == nullptr' failed.
unknown location(0): fatal error in "<suite>/<case>":
    signal: SIGABRT (application abort requested)
```

The first suite to fail is `amount_tests/GetFeeTest` — `ECC_Start`
is being re-entered while `secp256k1_context_sign` is still non-null
from a previous fixture. The `BasicTestingSetup` (or one of its
derived fixtures) appears to be missing the matching `ECC_Stop` on
teardown, so the next fixture's `ECC_Start` trips the assertion.

#### Numbers

Same docker container, same configure flags (`--enable-tests
--enable-wallet --with-incompatible-bdb`), same build:

| Branch | Full-suite failures |
|---|---|
| `upstream/develop` (`989ec2e1d`, no EVM changes) | **318** |
| This PR (`b69fb94c7`) | **434** = 318 baseline + 116 EVM cases that abort downstream of the same corruption |

The 116 EVM cases shipped in this PR pass **100% (796/796
assertions)** when run as a group via:

```
$ src/test/test_raptoreum --run_test=evm_smoke_tests:evm_evmtx_tests:\
  evm_state_tests:evm_host_tests:evm_apply_tests:evm_hashing_tests:\
  evm_process_tests:evm_connectblock_tests:evm_undo_tests:\
  evm_parallel_tests:evm_rlp_tests
*** No errors detected
Running 116 test cases...
  116 test cases out of 478 passed
  362 test cases out of 478 skipped
  796 assertions out of 796 passed
```

They also pass individually via `--run_test=<suite>` for each of
the 11 EVM suites. Their "failure" in the full-suite run is the
ECC_Start assertion blowing up the BasicTestingSetup ctor — same
pattern that hits every other downstream suite on `develop` today.

#### Recommendation

Keep this PR scoped to EVM. The fixture bug deserves its own
small PR — it's CI-hygiene work independent of the EVM
integration, and the team probably wants it merged on its own
timeline.

The likely fix is one of:

- Add a matching `ECC_Stop` to the fixture's teardown path.
- Replace per-fixture `ECC_Start`/`ECC_Stop` with a single
  process-wide `ECCVerifyHandle` (the upstream-Bitcoin-Core
  pattern).
- Switch the fixture to a `boost::test_tools::output_test_stream`
  global-fixture pattern.

Happy to open a follow-up PR with whichever shape the team
prefers, or to leave it for whoever owns the CI test
infrastructure.

#### Why this likely hasn't bitten in CI

The repo's `.github/workflows/build.yaml` runs `make check`,
which invokes the suite as a single binary, so it should hit
this. Either the CI is masking the failures (exit code from the
boost test runner being ignored?), or full-suite runs aren't
gating CI today. Worth a quick check on the CI side too.

If I'm misreading something — or the team has a different
expected invocation pattern for the test binary — happy to
adjust.

JSanchezFDZ:
Follow-up on the test-suite numbers I posted earlier:

I tracked down where the "318 failures across the suite" actually come from. They are **not** caused by the EVM integration. The cascade originates in `amount_tests/GetFeeTest`, where the line

```cpp
CFeeRate(MAX_MONEY, std::numeric_limits<size_t>::max() >> 1).GetFeePerK();
```

(inherited verbatim from Bitcoin/Dash) overflows `int64_t` inside `CFeeRate`'s two-arg constructor on Raptoreum, because Raptoreum's `MAX_MONEY` is 1000x larger than Bitcoin's. On modern toolchains the signed-overflow trap raises `SIGABRT`, Boost.test catches the signal but unwinds without running `BasicTestingSetup`'s destructor, so `ECC_Stop` is never called and every subsequent fixture constructor trips `assert(secp256k1_context_sign == nullptr)` in `ECC_Start()`.

I've opened #444 with the minimal fix (one line, plus comment). With that PR applied on top of `develop`, the full Boost test suite passes cleanly (362/362), including the 11 new EVM suites added in this PR.

Sequencing suggestion: merge #444 first, then I'll rebase this PR onto `develop` and the test report becomes trivially honest.

While confirming the fix I also noticed a second, unrelated build break on current `develop`: PR #436 added three new wallet-using RPC functions to `rpc/rpcassets.cpp` (`listassetsbalance`, `listunspentassets`, and the `mine=true` branch of `listassets`) without `#ifdef ENABLE_WALLET` guards, so `--disable-wallet` builds currently fail to compile. Opening a separate small PR for that as well.

JSanchezFDZ:
Follow-up to the comment above: while verifying the test-suite fix end-to-end I also tracked down two additional pre-existing breakages on `develop` that block `--disable-wallet` builds, and opened separate small PRs for each:

- **#446** — `rpc/rpcassets.cpp` lacks `#ifdef ENABLE_WALLET` guards around three new RPC handlers added in #436. `--disable-wallet` builds currently fail at compile time on that TU.

- **#447** — `configure.ac` has two copy-paste typos in the NAT-PMP detection block (one inverted branch, one stray character in a string literal). Result: `USE_NATPMP` is `AC_DEFINE`'d even when disabled, so `mapport.cpp` compiles in the NAT-PMP code path while `NATPMP_LIBS` stays empty — the daemon then fails to link with undefined references against `readnatpmpresponseorretry`, `sendnewportmappingrequest`, `sendpublicaddressrequest`.

With #444 + #446 + #447 all applied on top of `develop`, the `--disable-wallet --disable-bench --without-gui` daemon builds cleanly and the resulting `raptoreumd` runs (verified locally inside the depends/-managed Ubuntu 22.04 / GCC 11.4 build environment). The full Boost test suite also passes 362/362.

No new code in any of those three PRs touches consensus or the EVM work in this PR — they're all pure infrastructure fixes that happened to be in the way.

JSanchezFDZ:
## Progress update — EVM is now usable on-chain end-to-end, plus the D4 Smart-Asset mirror

Since the draft snapshot the branch has moved from "Phases 0–5 scaffolding" to a **fully on-chain-usable native EVM**, and the flagship differentiator — the **D4 Smart-Asset bidirectional mirror** — is implemented, tested, and proven e2e on a live regtest daemon. This is a heads-up for reviewers on what changed and, importantly, where I deliberately stopped and **why those parts want the audit, not more solo commits**.

Docs are the source of truth and have been kept current: [`docs/evm/STATUS.md`](docs/evm/STATUS.md) (phase/commit map), [`docs/evm/SMART-ASSET-MIRROR.md`](docs/evm/SMART-ASSET-MIRROR.md) (the D4 design-of-record), [`docs/evm/FUP.md`](docs/evm/FUP.md) (the follow-up register, with ✅-shipped markers), [`docs/evm/RPC.md`](docs/evm/RPC.md), [`docs/evm/PRECOMPILES.md`](docs/evm/PRECOMPILES.md).

### 1. D2 — EVM commitment hard fork (RIP-vote-gated)

The five EVM consensus roots are committed in the **coinbase special tx (`CCbTx` v3)** — `evmStateRoot`, `evmReceiptsRoot`, `evmBaseFee`, `evmGasUsed`, `evmExecTime` — gated by `EUpdate::EVM_COMMIT` (force-active at height 0 on regtest; **unregistered on testnet/mainnet** — those activate via the standard RIP miner+smartnode vote, configured when scheduled). Miner output == validator recompute via a single shared helper (`ComputeCoinbaseEvmCommitment`), so the miner can never build a template the validator would reject. EIP-1559 base-fee dynamics + burn/tip are live. Regression-locked in `evm_d2_consensus_tests` + `cbtx_evm_commit_tests`.

### 2. Value bridges both ways (UTXO ↔ EVM)

- `TRANSACTION_EVM_FUND` (UTXO → EVM) + `evm_fund` RPC — supply-conserving (the funded amount leaves the miner-claimable fee and reappears as EVM balance).
- `TRANSACTION_EVM_SPEND` (EVM → UTXO) — credits realised as coinbase outputs, verified by `CheckCoinbaseRealisesSpendCredits`; the miner now realises them (a real bug fix — a SPEND-bearing block would otherwise wedge mining).
- **Funded e2e proven** in `eth_funded_e2e_tests`: FUND → EIP-155-signed `eth_sendRawTransaction` → mine → `eth_getTransactionReceipt` status `0x1`, resolved by the Ethereum tx hash.

### 3. D4 — Smart-Asset bidirectional mirror (the flagship)

Every Smart Asset is callable as a standard ERC-20 at its per-asset precompile address, and units move **both ways** across the UTXO ↔ EVM boundary, supply-conserving and reorg-safe. Full design in [`SMART-ASSET-MIRROR.md`](docs/evm/SMART-ASSET-MIRROR.md). Key design choices reviewers should sanity-check:

- **No vault, no custody, no coinbase-asset channel.** wrap/unwrap reuse the per-tx asset-conservation **exemption** (the same one `MINT` uses) and each re-imposes a **constrained** conservation in its own `CheckSpecialTx` handler: for the declared `assetId` the (vout−vin) delta must be exactly ∓amount, and **every other asset must be balanced** — so the exemption can't be abused to mint/burn arbitrary assets. This is the security-critical guard.
- **EVM-side ledger** in the precompile's own storage trie (Solidity-compatible layout), so it's committed in `evmStateRoot` and reverted by the EVM reorg undo journal like any contract storage.
- **Invariant:** `sum(UTXO asset balances) + EVM wrappedSupply == circulatingSupply` (untouched).
- unwrap's EVM-side debit is the **authorization** for the UTXO mint: a shortfall rejects the whole block, so a minted output can never stand without its matching EVM burn.

Milestones M1–M6: the EVM ERC-20 ledger, the shared `asset_ledger` layout module, the wrap/unwrap consensus bridge, the **T-mirror convergence gate** (2000-op randomized sequence asserting both invariants after every op), wallet RPCs (`wrap_asset`/`unwrap_asset`/`get_asset_evm_address`), and an in-process consensus test of the constrained-delta guard (incl. a "silently burn another asset" attack). **Proven e2e on a live regtest daemon**: create+mint → wrap → `eth_call balanceOf` → unwrap → balances/supply verified; EVM ledger persists across daemon restart.

### 4. eth_* / asset correctness pass (each with a regression test)

A wallet-compatibility + correctness sweep — see the ✅-shipped rows in [`FUP.md`](docs/evm/FUP.md):

- **`eth_estimateGas`** now returns a complete tx gas limit (intrinsic 21000 + EIP-2028 calldata + binary-searched minimum execution gas honouring the 63/64 rule). It previously returned execution gas only, under-funding real txs.
- **EIP-1559 RPC surface** un-staled now the base fee is live: `eth_gasPrice` = next-block base fee + tip; `eth_maxPriorityFeePerGas` = 1 gwei; the block object now carries `baseFeePerGas`/`gasUsed`/`stateRoot`/`receiptsRoot` from the CCbTx v3 commitment, and real `difficulty`/`totalDifficulty`.
- **Tx-hash identity unified** across block listing / receipt / `eth_getTransactionByHash` (so dApps cross-reference block ↔ receipt); `eth_getTransactionByHash` now returns the full tx shape.
- **Legacy EIP-155 signing** primitive + a `"type":"0x0"` path in `evm_signTransaction`/`evm_sendTransaction`.
- **Asset resolver** O(N)→O(log N): `CAssetsCache::ResolveAssetIdByTag` with a **self-correcting verified hint index** — every hit re-verified against `mapAsset` with a scan fallback, so the result is a pure function of `mapAsset` and a stale hint can never diverge consensus (property + stale-hint tests).

### 5. CI

The Linux `evm-consensus-gate` job is a hard gate running Capa B (official Ethereum tests, ~99.95% of applicable Cancun, pass-floor + classified non-fixable allow-list — see [`OFFICIAL-TESTS.md`](docs/evm/OFFICIAL-TESTS.md)) + reorg fuzz + the D4 mirror + value-bridge + D2-commitment suites. Full local suite is green (572 cases).

### 6. Deliberately NOT done solo — these want the audit, not autonomous commits

I drew a hard line at any change that could **diverge consensus** without a safety-by-construction property or external review. Tracked in [`FUP.md`](docs/evm/FUP.md):

- **EIP-2930 access lists** in `eth_sendRawTransaction` (gas accounting must be byte-identical across nodes — fork risk if wrong).
- **EVM validation in `TestBlockValidity`/`VerifyDB`** (FUP-2.7): for miner-built blocks it's a tautology and doubles EVM exec per template; the `VerifyDB` half needs historical EVM state that isn't snapshotted. `ConnectTip` already validates, so the gap is defense-in-depth, not a hole.
- **Precompile encoding changes** (IPv6 in the MN struct, exact ChainLocks water-line) — they change returned bytes / determinism.
- True parallel EVM execution, releasing `cs_main` during execution, gas-cost calibration — large/data-dependent.

The genuinely next step for these — and for testnet activation — is **public testnet + a paid security audit** of the consensus-level EVM integration before mainnet. Happy to prep any of the deferred items as **separate, clearly-marked "pending-audit-review" branches** if the team wants to look at a specific one.

Review can proceed phase-by-phase against `STATUS.md`; the per-phase implementation notes live in the commit bodies. Questions still owed by the core team are in [`QUESTIONS.md`](docs/evm/QUESTIONS.md).

JSanchezFDZ:
## Review entry point + proposed decisions

To make this branch actually reviewable, the docs now include:

- **[`docs/evm/REVIEW-GUIDE.md`](docs/evm/REVIEW-GUIDE.md)** — start here. Subsystem-ordered reading path (not commit-by-commit), the 8 consensus hot-spots ranked with concrete "try to break this" prompts, and how to reproduce every verification claim locally (unit gates, Capa B, the live regtest e2e). The single most important property is spelled out and verifiable first: **merging this changes nothing on mainnet or testnet** — every consensus-visible behavior is gated on `EVM`/`EVM_COMMIT`, which are registered only for regtest.
- **[`docs/evm/QUESTIONS.md`](docs/evm/QUESTIONS.md)** — every open question now carries a **proposed answer** written so a single 👍 ratifies it. Q-A1 is marked resolved per the out-of-band clarification (the requirement — stay inside the existing dual-consensus and signature model, no invasive serialization changes — is satisfied by the special-tx wrapper design; details in the entry). Q-A2/Q-A6 are answered-by-implementation and only need ratification.
- **[`docs/evm/TESTNET-ACTIVATION.md`](docs/evm/TESTNET-ACTIVATION.md)** — concrete activation proposal (forced height on testnet ≈ tip+30d, bits 3/4, draft chainparams patch NOT applied, rollout + post-activation smoke checklist, abort lever). Nothing activates anywhere without sign-off on this document.

One item worth highlighting because it came from your requirements: assets are non-EVM **by default** (nothing touches the EVM unless explicitly wrapped), and **FUP-4.10** proposes a per-asset opt-out flag so an issuer can forbid wrapping at consensus level — a hard guarantee that a given asset can never acquire Solidity-risk exposure. Small change inside the already-tested wrap surface; awaiting your nod before implementing.

Happy to do the call whenever suits — the review guide doubles as the agenda.

JSanchezFDZ:
@BlockReq-AmyPan — agreed, `eth_getLogs` is exactly the surface indexers trip over. All five points are implemented and covered by a regression test that runs against a real mined chain (`eth_funded_e2e_tests::getlogs_filters_ordering_and_reorg`), plus a structural set in `eth_rpc_conformance_tests`.

Your two filter shapes are asserted directly:

- `{"fromBlock","toBlock","address":["0x..."],"topics":[null,["0x..","0x.."]]}` — address array + `[null, or-set]` positional topics; the test emits a `LOG2(0x..aa, 0x..bb)` and asserts only the log whose position-1 topic is in the or-set comes back.
- `{"blockHash":"0x...","topics":[]}` — empty topics = no constraint; returns the whole block's logs.

Point by point:

- **reject `blockHash` when `fromBlock`/`toBlock` is present** — done (EIP-234); passing both is a clean JSON-RPC error, not a silently-ignored field. Unknown or reorged-out hashes also error rather than returning `[]`.
- **order by `(blockNumber, transactionIndex, logIndex)`** — iteration is height-ascending → vtx-index-ascending → log-position-ascending. I also fixed a latent bug your wording caught: `logIndex` is now the log's position **within the block** (cumulative across its txs), not per-receipt. Two log-emitting txs in one block now report `logIndex` 0x0, 0x1, 0x2 (a per-receipt counter would have emitted 0x0, 0x1, 0x0 and broken every indexer). The test pins this.
- **`DisconnectTip` + reconnect consistency** — the test invalidates the log-bearing block, asserts its logs vanish from range queries and error by hash, then reconnects and asserts the exact same three logs return. There's also a guard so a receipt re-keyed to a different block after a reorg can never be emitted under the wrong block's coordinates (`receipt.blockHash` is cross-checked against the block being scanned).
- **explicit range-limit error instead of timeout** — an oversized *requested* numeric range fails fast with an explicit error **before** any chain access; the 10k-block cap is enforced. A numeric `toBlock` past the tip clamps to the tip (geth-style) instead of erroring.

Docs updated in [`docs/evm/RPC.md`](docs/evm/RPC.md). Thanks — this is precisely the kind of edge-case feedback that's most useful before testnet; if there are other RPCs you'd want a regression set for (I'd guess `eth_getTransactionReceipt` log bloom / `eth_getBlockByNumber` fullTx), name them and I'll add them the same way.

JSanchezFDZ:
## Pre-activation bank-grade review of the EVM consensus paths

I did a focused, independent review of the EVM consensus paths before any
activation is scheduled, reading each path end to end rather than relying on
tests. Full write-up with reproducible `file:line` evidence and fix designs is
in `docs/evm/BANK-GRADE-FINDINGS.md` (added in 5e0894c). Summary below.

**Framing / safety caveat first:** every EVM path is gated behind
`IsEvmActive` and the `EVM` deployment is **unregistered on mainnet and
testnet** (force-active at height 0 on regtest only). So **none of these is a
live-network issue today** — on any shared network the txs are rejected at
`CheckEvmCommon` with `evm-not-activated`. They are pre-merge/pre-activation
blockers, not live incidents. Because no EVM tx has ever been committed on a
shared network, fixing the serialization-level items now has zero
backward-compat cost.

### Findings

- **EVM-01 — CRITICAL — consensus trusts an unauthenticated `senderHash`.**
  The Ethereum signature is verified only at `eth_sendRawTransaction` ingest and
  then discarded; the serialized payload (`CEvmDeployTx`/`CEvmCallTx`/
  `CEvmSpendTx` in `src/evm/evmtx.h`) carries `senderHash`/`fromAddress` + nonce
  but **no r/s/v**. The consensus `Check*` functions (`src/evm/evmtx.cpp`)
  validate structure only, and `PreFlightCallOrDeploy`
  (`src/evm/process.cpp:130`) debits the payload-supplied account after checking
  only nonce + balance. Net: once the gate is open on a network with any
  non-cooperating node, a hand-built `EVM_SPEND` with `fromAddress = victim`
  drains any EVM account into a UTXO the attacker controls — no signature
  required. Fix design in the doc (carry the standard signed envelope and
  re-verify `sender == senderHash` in `Check` using the existing
  `DecodeRawEthTx`; the wallet-authored bridge txs — SPEND/FUND/WRAP/UNWRAP —
  need their own authorization envelope, which is the part I'd like the core
  team to weigh in on before I implement, since it touches consensus
  serialization).

- **EVM-02 — HIGH — chain-id split.** Ingest/signing uses `ActiveEvmChainId()`
  (regtest 7375, testnet 7374, mainnet 7373) but the execution context hardcodes
  `7373` (`src/evm/connectblock.cpp:263`, `src/validation.cpp:2597`). So on
  regtest/testnet a tx is signed for one id and the `CHAINID` opcode reports
  another. Also a prerequisite for the EVM-01 fix (recovery must use the signing
  id). Fix: parameterize via `Consensus::Params` (the existing FUP-1).

- **EVM-05 — HIGH — EVM state not undone on `VerifyDB` / `RollbackBlock`.** The
  EVM undo (`evm::ApplyUndoToDB`) runs only in `DisconnectTip`
  (`src/validation.cpp:3316`); `VerifyDB` (`:5344`) and `RollbackBlock`
  (`:5475`) disconnect via `DisconnectBlock` directly, which reverts UTXO/asset
  state but not EVM state. A `verifychain`-level-3+ over a range containing EVM
  txs re-applies them on un-rolled-back state → state-root divergence. Fix: move
  the EVM undo into a helper both disconnect paths call.

- **EVM-06 — MEDIUM (by design) — `value` is `uint64_t`.** Values ≥ 2⁶⁴ wei are
  unrepresentable. Flagged so the team can decide on widening to `uint256`
  before mainnet; not a safety bug.

- **DET-01 — open — determinism of the consensus-state-reading precompiles**
  (masternodes/chainlocks/llmq oracle) was not independently confirmed in this
  pass and is called out honestly rather than assumed safe.

### Shipped in this commit (5e0894c)

- **FUP-X.2** — `--disable-wallet` build fix: `evm_fund`/`wrap_asset`/
  `unwrap_asset` referenced `HELP_REQUIRING_PASSPHRASE` (wallet-only) in their
  unconditionally-compiled help blocks. Inlined a wallet-independent literal so
  wallet-less mining/relay nodes compile (the bodies already throw
  `RPC_METHOD_NOT_FOUND` without wallet support).

I deliberately did **not** land the EVM-01/EVM-02 consensus changes here — they
change EVM payload serialization / execution context and should be ratified as a
reviewed batch rather than pushed unilaterally. Designs are ready to implement on
green-light.

PR #444, opened 2026-05-12
test(amount): fix GetFeeTest int64 overflow causing 318-failure cascade

JSanchezFDZ:
## Summary

`amount_tests/GetFeeTest` aborts in current `develop` because the final boundary check feeds Raptoreum's `MAX_MONEY` (21B * COIN = 2.1e18) into `CFeeRate`'s two-arg constructor. The constructor multiplies by 1000 internally, producing 2.1e21, which overflows `int64_t` (max ~9.2e18). On modern toolchains the signed-overflow trap raises `SIGABRT`, Boost.test catches the signal but unwinds without running `BasicTestingSetup`'s destructor, so `ECC_Stop()` is never called. Every subsequent test case's fixture constructor then trips `assert(secp256k1_context_sign == nullptr)` inside `ECC_Start()`, cascading into hundreds of follow-on failures.

The test was inherited verbatim from Bitcoin/Dash, where `MAX_MONEY` is 21M * COIN (1000x smaller) and `MAX_MONEY * 1000` fits comfortably in `int64_t`. The 1000x bump to Raptoreum's cap predated this test; the boundary case just never got revisited.

Fix: use `OLD_MAX_MONEY` (= 21M * COIN, the Bitcoin-equivalent value already defined in `amount.h`) for that one line. The test still exercises the same "huge `nBytes`, should not crash" code path it always did, without triggering UB.

## Before

```
$ ./src/test/test_raptoreum
Running 362 test cases...
[...amount_tests/GetFeeTest aborts at line 82...]
test_raptoreum: key.cpp:341: void ECC_Start():
  Assertion `secp256k1_context_sign == nullptr' failed.
[...318 cascade failures across the rest of the suite...]
*** 318 failures are detected in the test module "Raptoreum Test Suite"
```

## After

```
$ ./src/test/test_raptoreum
Running 362 test cases...
*** No errors detected
```

## Why this matters

The cascade hides every other regression in the suite. Once `amount_tests` aborts, anyone running the unit tests sees a sea of "secp256k1 context already initialised" assertions and can't tell whether unrelated suites pass or fail. This is the kind of test infrastructure bug that quietly degrades the signal of CI over time.

Confirmed locally on `develop` (commit 989ec2e1d) inside the project's `depends/`-based Ubuntu 22.04 + GCC 11.4 build environment.

## Test plan

- [x] Build `test/test_raptoreum` against `develop` HEAD with the patch.
- [x] Run `./src/test/test_raptoreum --run_test=amount_tests`: 4/4 cases pass.
- [x] Run `./src/test/test_raptoreum`: 362/362 cases pass (UTF-8 locale; without it `fs_tests/fsbridge_fstream` fails on a separate locale-related issue, unrelated to this change).
- [x] Inspect diff: one line changed (plus comment explaining the why), no consensus code touched.

## Note for reviewers

This is the minimal localized fix. The deeper question -- whether `CFeeRate` should be hardened to accept fee amounts up to Raptoreum's full `MAX_MONEY` without overflowing -- is a wider design conversation (e.g. switch internal arithmetic to `__int128` or `boost::multiprecision::int128_t`, similar to how `CAmount128` is already used elsewhere in the codebase for asset balances). I'd be happy to open a separate issue for that if you'd like, but felt the boundary-case test fix should land independently so the suite stops hiding other regressions in the meantime.

### DIFF
--- a/src/test/amount_tests.cpp
+++ b/src/test/amount_tests.cpp
@@ -78,8 +78,14 @@ BOOST_AUTO_TEST_CASE(GetFeeTest)
         // some more integer checks
         BOOST_CHECK(CFeeRate(CAmount(26), 789) == CFeeRate(32));
         BOOST_CHECK(CFeeRate(CAmount(27), 789) == CFeeRate(34));
-        // Maximum size in bytes, should not crash
-        CFeeRate(MAX_MONEY, std::numeric_limits<size_t>::max() >> 1).GetFeePerK();
+        // Maximum size in bytes, should not crash.
+        // NB: use OLD_MAX_MONEY (== Bitcoin's MAX_MONEY) rather than Raptoreum's
+        // MAX_MONEY (1000x larger). With Raptoreum's 21B-coin cap, MAX_MONEY * 1000
+        // overflows int64_t inside CFeeRate's constructor (2.1e21 > int64_t::max).
+        // The overflow is undefined behaviour and aborts under signed-overflow
+        // sanitizers / -ftrapv builds, which in turn skips the fixture destructor
+        // and cascades into the rest of the suite (ECC_Start assertion).
+        CFeeRate(OLD_MAX_MONEY, std::numeric_limits<size_t>::max() >> 1).GetFeePerK();
         }
 
 BOOST_AUTO_TEST_CASE(BinaryOperatorTest)


npq7721 (approved):
LGTM

Issue #445, opened 2026-05-12
CFeeRate internal arithmetic overflows int64_t for fees near MAX_MONEY


JSanchezFDZ:
## Summary

`CFeeRate`'s two-arg constructor and `GetFee()` perform internal arithmetic that overflows `int64_t` for fee amounts anywhere near Raptoreum's `MAX_MONEY` (21B * COIN = 2.1e18). The constructor in `src/policy/feerate.cpp` computes:

```cpp
CFeeRate::CFeeRate(const CAmount &nFeePaid, size_t nBytes_) {
    ...
    if (nSize > 0)
        nSatoshisPerK = nFeePaid * 1000 / nSize;   // <- overflows for nFeePaid > int64_t::max / 1000
    ...
}
```

`int64_t::max() / 1000 ≈ 9.2e15`, which is ~228x smaller than `MAX_MONEY`. So any fee amount within roughly the upper 99.5% of the valid `CAmount` range overflows here, with undefined behaviour. Under signed-overflow sanitizers / `-ftrapv` builds, this aborts the process.

This was inherited from upstream Bitcoin, where `MAX_MONEY = 21M * COIN ≈ 2.1e15`, and `MAX_MONEY * 1000 ≈ 2.1e18` fits comfortably in `int64_t`. The 1000x bump to Raptoreum's cap didn't get reflected in `CFeeRate`.

Real-world impact today is small (no caller passes single-tx fees that large), but:

- It surfaced as a test-suite cascade — see PR #444, which fixes the specific test boundary case by switching to `OLD_MAX_MONEY`. That's a minimal workaround, not a real fix to `CFeeRate` itself.
- Defensive arithmetic for consensus-adjacent code is cheap; UB lurking near `MAX_MONEY` is exactly the kind of thing that becomes a footgun later (e.g. when fuzz-testing or sanitizer CI lands).

## Proposed fix

Promote internal arithmetic in `CFeeRate::CFeeRate(CAmount, size_t)` and `CFeeRate::GetFee(size_t)` to 128-bit. Two natural options:

1. **`__int128`** — zero dependencies, supported by GCC/Clang on every platform Raptoreum already targets. Cheapest patch.
2. **`boost::multiprecision::int128_t` / `CAmount128`** — already used in the codebase for asset balances (`src/amount.h:15`). More portable and consistent with existing conventions.

[...]

PR #446, opened 2026-05-12
build: add ENABLE_WALLET guards to rpc/rpcassets.cpp

JSanchezFDZ:
## Summary

Restore the ability to build with `./configure --disable-wallet` on current `develop`.

PR #436 (`Bug/createrawtransaction for asset`) added three new RPC handlers in `src/rpc/rpcassets.cpp` that unconditionally reference wallet-only symbols (`CWallet`, `GetWalletForJSONRPCRequest`, `COutput`, `CCoinControl`, `CInputCoin`, `CoinType`):

- `listassetsbalance` (lines 815-869)
- `listunspentassets` (lines 871-1064)
- `listassets`, `mine=true` branch (lines 1160-1203)

The RPC command table at the bottom of the file already wraps the registrations for `listassetsbalance` and `listunspentassets` in `#ifdef ENABLE_WALLET`, but the function bodies themselves were left unguarded. Since the file's includes for wallet headers are gated on `ENABLE_WALLET` (lines 16-20), `--disable-wallet` builds fail to compile this TU with dozens of "`CWallet` was not declared in this scope" errors.

## Fix

- Wrap the two wallet-only function bodies (`listassetsbalance`, `listunspentassets`) in `#ifdef ENABLE_WALLET` / `#endif`. The matching table entries are already guarded, so when the bodies are absent the function pointers are simply never referenced.
- In `listassets`, the function is intentionally mixed: `mine=false` is wallet-independent (queries `passetsdb` only), `mine=true` needs the wallet. Guard only the `mine=true` branch, and in the `#else` path throw `RPC_METHOD_NOT_FOUND` so a client that requests `mine=true` against a no-wallet node sees a clear error instead of silently getting wrong data.

## Test plan

- [x] `./configure --disable-bench --without-gui --disable-wallet`, then `make rpc/libraptoreum_server_a-rpcassets.o` — compiles cleanly (was broken on `develop`).
- [x] `nm` on the resulting `.o` confirms zero unresolved `CWallet` / `GetWalletForJSONRPCRequest` symbols.
- [x] `./configure ... --enable-wallet --with-incompatible-bdb`, then recompile the same TU — still builds.

Note: a separate, pre-existing link-time failure in `src/mapport.cpp` (undefined references to `readnatpmpresponseorretry` / `sendnewportmappingrequest` from `libnatpmp`) keeps the no-wallet `raptoreumd` link from completing on this build host. That's an unrelated NAT-PMP library-ordering issue and not in scope here.

## Why this matters

- CI builds with `--disable-wallet` (e.g. for testing headless/light-node configurations) are currently broken on `develop`.
- Future Guix-style deterministic builds typically need both `--enable-wallet` and `--disable-wallet` paths to work.

Related: #443 (EVM integration, runs into this same restriction during test-suite verification), #444 (the `amount_tests` overflow fix that lets the test suite report honest pass/fail counts), #445 (broader `CFeeRate` hardening tracking).

### DIFF
--- a/src/rpc/rpcassets.cpp
+++ b/src/rpc/rpcassets.cpp
@@ -812,6 +812,7 @@ UniValue getassetdetailsbyid(const JSONRPCRequest &request) {
     return result;
 }
 
+#ifdef ENABLE_WALLET
 UniValue listassetsbalance(const JSONRPCRequest &request) {
     if (request.fHelp || !Updates().IsAssetsActive(::ChainActive().Tip()) || request.params.size() > 0)
         throw std::runtime_error(
@@ -1062,6 +1063,7 @@ UniValue listunspentassets(const JSONRPCRequest& request)
 
     return results;
 }
+#endif // ENABLE_WALLET
 
 UniValue listassets(const JSONRPCRequest &request) {
     RPCHelpMan{"listassets",
@@ -1158,6 +1160,7 @@ UniValue listassets(const JSONRPCRequest &request) {
     UniValue result(UniValue::VOBJ);
 
     if (mine) {
+#ifdef ENABLE_WALLET
         // Get only assets owned by this wallet
         std::shared_ptr<CWallet> const wallet = GetWalletForJSONRPCRequest(request);
         if (!wallet) {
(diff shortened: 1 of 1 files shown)


npq7721 (approved):
LGMT

Issue #448, opened 2026-06-19
Setgenerate false reponse mentions incorrect number of threads.


Leonardiae:
After setgenerate false the response is always the number of threads from the first time I started. I reproduced this in 2.0.3 and 2.0.4.0.1.

```
11:55:35 setgenerate true 8
11:55:35 8 of 24
11:58:33 setgenerate false
11:58:33 8 of 24
11:58:38 setgenerate true 20
11:58:38 20 of 24
11:59:40 setgenerate false
11:59:40 8 of 24
```

PR #449, opened 2026-06-19
Update mining.cpp

Leonardiae:
Fix for https://github.com/Raptor3um/raptoreum/issues/448

### DIFF
--- a/src/rpc/mining.cpp
+++ b/src/rpc/mining.cpp
@@ -1219,8 +1219,8 @@ UniValue setgenerate(const JSONRPCRequest &request) {
     }
 
 
-    gArgs.ForceSetArg("-gen", (fGenerate ? "1" : "0"));
-    gArgs.ForceSetArg("-genproclimit", itostr(nGenProcLimit));
+    gArgs.SoftSetArg("-gen", (fGenerate ? "1" : "0"));
+    gArgs.SoftSetArg("-genproclimit", itostr(nGenProcLimit));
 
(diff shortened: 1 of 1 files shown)


CodeIsTheKey (approved):
This looks good to me!

JSanchezFDZ (approved):
Reviewed `c9d5f0316c1d0d66beb54758c4da4d7007780916`, the complete response-only change and the full `setgenerate`/`GenerateRaptoreums` path against the release target. **Approved: no new defect found.**

The response now reports `0 of N` when generation is stopped. The previous text could report the requested/configured thread count despite creating no mining threads. The existing string format and miner call/arguments are unchanged.

I traced disabled generation, a zero limit, negative limits, positive limits, and the omitted/configured default. They agree with the new expression: disabled/zero → 0; negative → available cores; positive → requested threads; omitted → existing configured limit, default 1. `GenerateRaptoreums` stops the prior thread group before its early return. The inspected flow also matches the fetched release branch outside this change.

**Validation:** complete diff/caller review, default and RPC/CLI registration checks, comparison with the release target, and `git diff --check` passed. These are source-level control-flow checks, not executed RPC cases. No daemon, mining worker, or full build was run.

Optional follow-up: add an assertion for the stop response and document the `active threads of available cores` result. Any integration test must account for the existing `setgenerate` rejection on networks with `MineBlocksOnDemand()`. No additional abstraction or performance change is needed.

PR #457, opened 2026-06-22
rpc: mine to the wallet selected on the setgenerate endpoint (#450)

### LINKED ISSUE #450: Raptoreum Core version v2.0.04.01-unk / OS : Windows / Command / setgenerate
Raptoreum Core version v2.0.04.01-unk : OS : Windows - When you're in console, select a wallet, use command setgenerate true x , the mining rewards go to default wallet, not to selected wallet. Very minor issue.

Image

Image

JSanchezFDZ:
## Problem

`setgenerate` always routed the coinbase rewards to the first loaded wallet
(`GetFirstWallet`), ignoring which wallet the call was made against. In a
multi-wallet setup the rewards therefore went to the default wallet instead of
the one selected in the GUI console (issue #450).

## Change

Thread the wallet name from the `/wallet/<name>` RPC endpoint down to the miner:

- `setgenerate` reads it with `GetWalletNameFromJSONRPCRequest` (no throw when no
  wallet is specified, so CLI/single-wallet use is unchanged),
- passes it to `GenerateRaptoreums(..., walletName)`,
- `RaptoreumMiner` resolves it with `GetWallet` and falls back to `GetFirstWallet`
  when the name is empty or the wallet is not loaded.

`GetWalletNameFromJSONRPCRequest` is now declared in `wallet/rpcwallet.h` so the
mining RPC can use it (guarded by `ENABLE_WALLET`).

## Testing

Compile-validated on `develop` with `--enable-wallet` (miner.cpp, rpc/mining.cpp,
wallet/rpcwallet.cpp all build cleanly). The actual reward routing was **not**
exercised at runtime: `setgenerate` is rejected on regtest (`MineBlocksOnDemand`)
and observing a real coinbase requires mining on a live network. A maintainer
with a mining setup can confirm by selecting a non-default wallet and checking
the coinbase destination. The change is backward compatible — with no wallet
endpoint the behaviour is identical to before.

Closes #450.

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -66,6 +66,12 @@ double nHashesPerSec = 0;
 uint64_t nHashesDone = 0;
 std::string alsoHashString;
 
+// Name of the wallet selected for mining rewards (issue #450). Set by
+// GenerateRaptoreums() before the miner threads start; empty means "use the
+// first loaded wallet" (previous behaviour). It is only written while the
+// miner is stopped, so the worker threads read a stable value.
+static std::string g_miningWalletName;
+
 int64_t UpdateTime(CBlockHeader *pblock, const Consensus::Params &consensusParams, const CBlockIndex *pindexPrev) {
     int64_t nOldTime = pblock->nTime;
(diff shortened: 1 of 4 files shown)


npq7721 (approved):
Good enough

PR #458, opened 2026-06-22
qt: label asset transactions by type in the history list (#382)

### LINKED ISSUE #382: Naming asset tx types in the GUI
Currently createasset and mintasset as well as asset sends are shown as "Transaction to yourself" in the transaction menu if going to an own wallet. 

It would be really nice if we could include the name of the asset as well as the transaction type create, mint,  send, future send etc...

JSanchezFDZ:
## Problem

createasset, mintasset and updateasset transactions appear in the wallet's
transaction list with a generic type — typically "Payment to yourself" when
they touch the user's own wallet — so they are indistinguishable from ordinary
sends (issue #382).

## Change

- Add `AssetCreate` / `AssetMint` / `AssetUpdate` record types.
- After `decomposeTransaction`, tag asset special transactions
  (`TRANSACTION_NEW_ASSET` / `TRANSACTION_MINT_ASSET` / `TRANSACTION_UPDATE_ASSET`)
  with the matching type. Amount decomposition is left untouched; only the
  displayed type label and colour change.
- Render them as "Create Asset" / "Mint Asset" / "Update Asset" in
  `formatTxType`, and colour them like other internal operations.

Asset *transfers* (ordinary transactions carrying asset outputs) are left as a
possible follow-up since they have no dedicated tx type to key off.

## Testing

Built with `--with-gui=qt5 --enable-wallet`: `transactionrecord.cpp`,
`transactiontablemodel.cpp` and the asset dialogs compile cleanly (uic/moc + CXX)
with the new enum values handled in the type switches.

Closes #382.

### DIFF
--- a/src/qt/transactionrecord.cpp
+++ b/src/qt/transactionrecord.cpp
@@ -334,6 +334,23 @@ TransactionRecord::decomposeTransaction(interfaces::Wallet &wallet, const interf
 
         // LogPrintf("TransactionRecord::%s TxId: %s, vOutIdx: %d, Unhandled\n", __func__, hash.ToString(), vOutIdx);
     }
+
+    // Tag asset special transactions with their operation so they show up as
+    // "Create/Mint/Update Asset" instead of a generic send/receive or
+    // "Payment to yourself" entry when they touch the user's own wallet (issue #382).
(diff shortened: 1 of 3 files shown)


npq7721 (approved):
LGTM

PR #461, opened 2026-08-24
smartnode: fix two crashes in `smartnode payments` and speed up fee computation

JSanchezFDZ:
Two fixes to `smartnode payments`, both in `src/rpc/smartnode.cpp`.

## 1. Crash on the genesis block

The block reward was derived from `GetBlockSubsidy(pindex->pprev->nBits, ...)`. For the genesis block `pindex->pprev` is null, so `smartnode payments <genesis-hash>` dereferenced null and crashed the whole node. Guarded with a clear JSON-RPC error.

## 2. Fee computation: crash on `-txindex=0`/pruning + slow

The block fees were recomputed by calling `GetTransaction()` for **every input of every transaction** in the block:

- **Latent crash:** on a node running with `-txindex=0` or pruning, `GetTransaction()` returns null for a confirmed prevout and the result was dereferenced without a null check (`txPrev->vout[...]`), taking the daemon down. (txindex is on by default, `DEFAULT_TXINDEX = true`, so this only affected non-default configs.)
- **Slow:** O(inputs) individual tx reads per block, times `count`.

Now the spent input values are read from the block's undo data (one sequential read per block), mirroring `getblock` verbosity 3 and `getblockstats`. This removes the crash and is much faster for per-block scanning.

## Testing

Built `raptoreumd` from this branch and verified on regtest:

- **Genesis:** unpatched → node crashes (`EOF reached`); patched → clean `-8` error, node stays up.
- **Fee loop, `-txindex=0`:** unpatched → `smartnode payments <block-with-a-transaction>` crashes the node; patched → returns and the node stays up.
- A functional regression test (`test/functional/rpc_smartnode_payments.py`) covers the genesis case.
- Full C++ unit test suite passes (no regression).

### DIFF
--- a/src/rpc/smartnode.cpp
+++ b/src/rpc/smartnode.cpp
@@ -485,7 +485,11 @@ UniValue smartnode_payments(const JSONRPCRequest &request) {
             payeesArr.push_back(obj);
         }
 
-        const auto dmnPayee = deterministicMNManager->GetListForBlock(pindex).GetMNPayee();
+        // The smartnode paid by a block is the payee of the list as it stood at the
+        // previous block, which is what GetBlockTxOuts() used to fill in the payees
+        // above. Using the list at this block instead would name the smartnode due
+        // to be paid by the *next* one.
+        const auto dmnPayee = deterministicMNManager->GetListForBlock(pindex->pprev).GetMNPayee();
         protxObj.pushKV("proTxHash", dmnPayee == nullptr ? "" : dmnPayee->proTxHash.ToString());
         protxObj.pushKV("amount", payedPerSmartnode);
         protxObj.pushKV("payees", payeesArr);
--- a/src/rpc/smartnode.cpp
+++ b/src/rpc/smartnode.cpp
@@ -404,6 +404,14 @@ UniValue smartnode_payments(const JSONRPCRequest &request) {
 
     while (vecPayments.size() < uint64_t(std::abs(nCount)) && pindex != nullptr) {
 
+        if (pindex->pprev == nullptr) {
+            // The genesis block has no previous block to derive the block reward
+            // from and pays no smartnode. Bail out with a clear error instead of
+            // dereferencing a null pprev below, which crashes the node.
+            throw JSONRPCError(RPC_INVALID_PARAMETER,
+                               "Smartnode payment information is not available for the genesis block");
+        }
+
         CBlock block;
         if (!ReadBlockFromDisk(block, pindex, Params().GetConsensus())) {
             throw JSONRPCError(RPC_INTERNAL_ERROR, "Can't read block from disk");
--- a/src/rpc/smartnode.cpp
+++ b/src/rpc/smartnode.cpp
@@ -427,20 +427,34 @@ UniValue smartnode_payments(const JSONRPCRequest &request) {
         // GetTransaction() lookup returned null for confirmed prevouts on such nodes and
         // the daemon crashed dereferencing it.
         CAmount nBlockFees{0};
-        if (IsBlockPruned(pindex)) {
-            throw JSONRPCError(RPC_MISC_ERROR, "Undo data not available (pruned data)");
-        }
         CBlockUndo blockUndo;
         if (!UndoReadFromDisk(blockUndo, pindex)) {
+            // The read itself is the source of truth for undo availability;
+            // IsBlockPruned() only refines the error message after a failure.
+            if (IsBlockPruned(pindex)) {
+                throw JSONRPCError(RPC_MISC_ERROR, "Undo data not available (pruned data)");
+            }
             throw JSONRPCError(RPC_INTERNAL_ERROR, "Can't read undo data from disk");
         }
         // vtx[0] is the coinbase (no inputs); blockUndo.vtxundo is indexed for all but the coinbase.
+        if (blockUndo.vtxundo.size() != block.vtx.size() - 1) {
+            throw JSONRPCError(RPC_INTERNAL_ERROR,
+                               strprintf("Undo data for block %s is inconsistent: %d transaction undo records for %d transactions",
+                                         pindex->GetBlockHash().ToString(), blockUndo.vtxundo.size(), block.vtx.size()));
+        }
         for (size_t i = 1; i < block.vtx.size(); ++i) {
+            const CTransaction &tx = *block.vtx[i];
+            const CTxUndo &txUndo = blockUndo.vtxundo[i - 1];
+            if (txUndo.vprevout.size() != tx.vin.size()) {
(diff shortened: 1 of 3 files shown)


src/rpc/smartnode.cpp
```diff
@@ -404,27 +405,42 @@ UniValue smartnode_payments(const JSONRPCRequest &request) {
...
+        // Besides being much faster, this also works on nodes without -txindex: the old
+        // GetTransaction() lookup returned null for confirmed prevouts on such nodes and
+        // the daemon crashed dereferencing it.
         CAmount nBlockFees{0};
-        NodeContext &node = EnsureNodeContext(request.context);
-        for (const auto &tx: block.vtx) {
-            if (tx->IsCoinBase()) {
-                continue;
-            }
+        if (IsBlockPruned(pindex)) {
```
npq7721: Should we rely on UndoReadFromDisk() as the source of truth for undo availability instead of IsBlockPruned(pindex)? The current check may be broader than necessary and could reject blocks that are otherwise readable, depending on node state

src/rpc/smartnode.cpp
```diff
@@ -404,27 +405,42 @@ UniValue smartnode_payments(const JSONRPCRequest &request) {
...
+        }
+        // vtx[0] is the coinbase (no inputs); blockUndo.vtxundo is indexed for all but the coinbase.
+        for (size_t i = 1; i < block.vtx.size(); ++i) {
             CAmount nValueIn{0};
-            for (const auto &txin: tx->vin) {
-                uint256 blockHashTmp;
-                CTransactionRef txPrev = GetTransaction(/* block_index */ nullptr, node.mempool, txin.prevout.hash,
-                                                                          Params().GetConsensus(), blockHashTmp);
-                nValueIn += txPrev->vout[txin.prevout.n].nValue;
+            for (const Coin &coin: blockUndo.vtxundo.at(i - 1).vprevout) {
```
npq7721: Could we validate that blockUndo.vtxundo.at(i - 1).vprevout matches the transaction’s input count before summing fees? Right now this assumes the undo data is complete and in sync with block.vtx[i]; if it isn’t, we’d silently compute the wrong fees instead of failing fast with a clearer error.

src/rpc/smartnode.cpp
```diff
@@ -404,27 +405,42 @@ UniValue smartnode_payments(const JSONRPCRequest &request) {
...
-            }
+        if (IsBlockPruned(pindex)) {
+            throw JSONRPCError(RPC_MISC_ERROR, "Undo data not available (pruned data)");
+        }
+        CBlockUndo blockUndo;
+        if (!UndoReadFromDisk(blockUndo, pindex)) {
+            throw JSONRPCError(RPC_INTERNAL_ERROR, "Can't read undo data from disk");
+        }
+        // vtx[0] is the coinbase (no inputs); blockUndo.vtxundo is indexed for all but the coinbase.
+        for (size_t i = 1; i < block.vtx.size(); ++i) {
```
npq7721, line 447: Please consider checking blockUndo.vtxundo.size() == block.vtx.size() - 1 before iterating. This loop assumes perfect alignment between the undo vector and block.vtx, and an explicit guard would make failures clearer

src/rpc/smartnode.cpp
JSanchezFDZ: Good point — reordered in 1afb323b2: `UndoReadFromDisk()` is now the source of truth, and `IsBlockPruned()` is only consulted after a failed read to refine the error message for pruned nodes. Worth noting the original order couldn't reject readable undo data — `IsBlockPruned()` requires `BLOCK_HAVE_DATA` to be missing, in which case the `ReadBlockFromDisk()` call above would already have thrown — but reading first is strictly cleaner.

src/rpc/smartnode.cpp
JSanchezFDZ: Added in 1afb323b2 — the loop is now preceded by a `vtxundo.size() == vtx.size() - 1` check that fails with an explicit error naming the block, mirroring the consistency check `DisconnectBlock()` already performs. This matches the writer invariant in `ConnectBlock()`, which pushes one `CTxUndo` for every non-coinbase transaction (including zero-input special transactions). FWIW, `getblockstats` inherits the same alignment assumption from upstream Bitcoin Core — happy to harden it the same way in a follow-up if you'd like.

src/rpc/smartnode.cpp
JSanchezFDZ: Added in 1afb323b2 — each transaction's undo record is now checked for `vprevout.size() == vin.size()` before summing, failing fast with the tx hash in the error instead of silently miscomputing fees (same guard `DisconnectBlock()` uses). In practice `UndoReadFromDisk()` already verifies a checksum keyed by the previous block hash, so a mismatch would indicate a writer-side bug rather than disk corruption — but failing loudly is clearly better.

JSanchezFDZ:
Pushed two more fixes to the same RPC. I found both while writing functional-test coverage for this PR, and I would rather they travel with it than land as a separate change to the same function — happy to split them out if you prefer.

**`count` was silently ignored.** The help documents `smartnode payments ( "blockhash" count )`, so a call that passes a count arrives with two parameters, but the guard read

```cpp
request.params.size() > 2 ? ParseInt64V(request.params[1], "count") : 1
```

which only holds when a third argument is supplied, and there is no third argument. So `smartnode payments <hash> -3` returned one block, not three. Changed to `> 1`.

**The reported `proTxHash` was one block out.** Each entry reports a `proTxHash` next to the `payees` and `amount` for that block, but it came from `GetListForBlock(pindex)` while the payees come from `GetBlockTxOuts()`, which uses `ChainActive()[nBlockHeight - 1]`. The two named different smartnodes: the `proTxHash` was the node due to be paid by the *next* block. It now reads from `pindex->pprev`, matching the payees in the same object. (The projection loop in `smartnode winners` confirms the convention: it uses the list at the tip for height tip+1.)

Verification, on regtest with three smartnodes:

| | matches winners[h] | matches winners[h+1] |
|---|---|---|
| before | 0 / 4 | 4 / 4 |
| after | 4 / 4 | 0 / 4 |

The `count` fix is covered in `rpc_smartnode_payments.py`; reverting it makes that test fail with `[5] == [3, 4, 5]`.

One caveat worth flagging: `rpc_smartnode_payments.py` **cannot currently run** on `develop`, and neither can any other functional test that mines a block. `TestNode.generate()` calls `get_deterministic_priv_key()`, which is not defined anywhere in the framework, so it gets dispatched to the node as an RPC and fails with "Method not found". I have opened #467 to fix that and six other breakages in the functional test framework. Until it lands, the regression test in this PR is not executable in CI — though CI does not run the functional suite at all today, which is why none of this was visible.

PR #462, opened 2026-08-25
llmq: fix out-of-bounds crash on DKG justification (off-by-one)

JSanchezFDZ:
## Summary

`PreVerifyMessage(CDKGJustification)` validated the contribution index with `p.first > members.size()`, which lets `p.first == members.size()` through. `ReceiveMessage()` then uses it as `members[p.first]` — an out-of-bounds access into the members vector — and dereferences the resulting garbage pointer (`member2->dmn->proTxHash`), crashing the node (SIGSEGV).

## Impact

The check runs in the DKG worker thread, so the crash is **not** caught by the P2P message handler's `try/catch`. A single malicious quorum member can send a crafted `QJUSTIFICATION` and crash other smartnodes during the justification phase of DKG. This is gated by `SPORK_17_QUORUM_DKG_ENABLED` (active on mainnet, since quorums back ChainLocks/InstantSend).

`src/llmq/quorums_dkgsession.cpp:756`:

```cpp
if (p.first > members.size()) {   // lets p.first == members.size() pass
```

used later at `:830` as `members[p.first]`.

## Fix

Reject `p.first >= members.size()`, matching the other DKG message handlers (complaints, premature commitments) which already bound-check, and matching Dash upstream.

## Testing

Verified the branch builds cleanly and the existing `evo`/`llmq` unit test suites pass (no regression). The change is a one-character bound fix (`>` → `>=`) with an explanatory comment.

### DIFF
--- a/src/llmq/quorums_dkgsession.cpp
+++ b/src/llmq/quorums_dkgsession.cpp
@@ -753,7 +753,10 @@ namespace llmq {
 
         std::set <size_t> contributionsSet;
         for (const auto &p: qj.contributions) {
-            if (p.first > members.size()) {
+            // p.first is used as an index into `members` in ReceiveMessage(); it must be
+            // strictly less than members.size(). Using '>' let p.first == members.size()
+            // through, causing an out-of-bounds access later. Reject it here.
+            if (p.first >= members.size()) {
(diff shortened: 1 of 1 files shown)


npq7721 (approved):
LGTM

PR #464, opened 2026-08-25
node: defensive hardening for latent crash paths + remove debug print

JSanchezFDZ:
## Summary

Defensive fixes for node-aborting conditions in RPC and P2P/consensus paths, plus removal of a stray debug print. These are mostly latent (hard to trigger today) but rely on fragile invariants; a couple are cheap wins.

## Changes

- **rpc/blockchain.cpp (`gettxout`)** — null-check `LookupBlockIndex(best block)` before dereferencing (every other `LookupBlockIndex` in the file is checked).
- **rpc/smartnode.cpp (`smartnodelist`)** — null-check `ChainActive()[nLastPaidHeight]` before reading `nTime`.
- **llmq/quorums_dkgsessionmgr.cpp** — reject a 1-byte `QCONTRIB`/`QPCOMMITMENT` instead of reading the second byte past the end of the buffer.
- **llmq/quorums.cpp (`QDATA` handler)** — null-check `quorumVvec` (a shared_ptr) before `size()`.
- **assets/`validateAmount`** — reject `decimalPoint > 8` (the `10^(8-decimalPoint)` divisor underflows to 0 → SIGFPE) and use an integer lookup table instead of `pow()`, keeping this off floating-point math in a consensus path. Unit test added.
- **rpc/rpcevo.cpp (`protx list`)** — remove a leftover `std::cout` debug line that ran on every call inside the `cs_main` critical section.

## Testing

Branch builds cleanly; `assets`/`evo` unit suites pass, and the new `validate_amount_decimal_point` unit test passes (covers the 0..8 range and asserts `decimalPoint > 8` is rejected instead of dividing by zero).

### DIFF
--- a/src/assets/assets.cpp
+++ b/src/assets/assets.cpp
@@ -586,7 +586,15 @@ bool GetAssetData(const CScript &script, CAssetOutputEntry &data) {
 }
 
 bool validateAmount(const CAmount nAmount, const uint16_t decimalPoint) {
-    if (nAmount % int64_t(pow(10, (8 - decimalPoint))) != 0) {
+    // Assets carry at most 8 decimals. A larger decimalPoint would make the divisor
+    // below (10^(8-decimalPoint)) underflow to 0 and divide by zero (SIGFPE). Using an
+    // integer lookup table also keeps this off floating-point math in a consensus path.
+    if (decimalPoint > 8) {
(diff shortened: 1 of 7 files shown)


npq7721 (approved):
LGTM

PR #466, opened 2026-08-31
ci: build and test on Ubuntu 26.04

JSanchezFDZ:
Follow-up to #460. That PR made the tree build with GCC 15 / Ubuntu 26.04, but nothing in CI actually produces Ubuntu 26 artifacts, so the fix is not exercised and there is nothing to ship for that platform.

This adds an Ubuntu 26 entry to the `build-linux` and `test-linux` matrices.

### Artifact naming

The platform id is `ubuntu26`, so `package-artifacts` emits names consistent with the existing Linux targets:

```
raptoreum-ubuntu26-<version>.tar.gz
raptoreum-ubuntu26-debug-<version>.tar.gz
raptoreum-ubuntu26-not_strip-<version>.tar.gz
```

and the uploaded artifact is `raptoreum-ubuntu26-<version>`, matching the `raptoreum-ubuntu22-*` / `raptoreum-ubuntu24-*` pattern the release flow already expects. Nothing about the packaging or checksum steps needed changing — they are all driven by `platform-id`.

### About the preview runner

`ubuntu-26.04` is still a **public preview** image (announced 2026-06-11): no SLA, and capacity is still being balanced. So both Ubuntu 26 jobs carry `continue-on-error`, driven by an `experimental: true` key on the matrix entry, so an outage on a preview runner cannot turn the whole workflow red for unrelated changes. The artifacts are still built and uploaded normally. Once the image goes GA, deleting the `experimental` key from the two entries is enough to make it a blocking target.

If you would rather have it blocking from day one, say so and I will drop the flag.

### Package list

Unchanged from the other Linux targets. I checked each package in an `ubuntu:26.04` image and they are all still present, `bsdmainutils` included.

### On signing

I looked at whether this could feed the signing step and I do not think it can be automated as things stand: the Windows flow in `SIGNING_QUICK_START.md` signs with a SafeNet USB token via `signtool`, and a physical token cannot be used from a GitHub-hosted runner. There is also no release/publish job in any workflow today — artifacts are downloaded and released by hand. So what I have done here is make sure the Ubuntu 26 artifacts come out under the exact naming convention the manual flow already relies on.

If the intent was something more than that — a self-hosted runner with the token attached, a move to cloud signing (Azure Trusted Signing or similar), or an actual `gh release` job wired into this workflow — tell me which and I will do it as a separate PR; those touch secrets and release policy, so I did not want to guess.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -69,6 +69,7 @@ jobs:
     name: ${{ matrix.platform.name }} Build
     needs: get-version
     runs-on: ${{ matrix.platform.runner }}
+    continue-on-error: ${{ matrix.platform.experimental || false }}
     strategy:
       fail-fast: false
       matrix:
@@ -86,6 +87,17 @@ jobs:
             packages: curl libcurl4-openssl-dev build-essential libtool autotools-dev automake pkg-config python3 bsdmainutils cmake ccache
             platform-id: ubuntu24
             strip-command: strip
+
+          # ubuntu-26.04 is still a public preview runner image (no SLA, capacity
+          # may be constrained), so this build is allowed to fail without failing
+          # the workflow. Drop continue-on-error once the image reaches GA.
+          - name: Ubuntu 26
+            runner: ubuntu-26.04
+            host: x86_64-pc-linux-gnu
+            packages: curl libcurl4-openssl-dev build-essential libtool autotools-dev automake pkg-config python3 bsdmainutils cmake ccache
+            platform-id: ubuntu26
+            strip-command: strip
+            experimental: true
 
           - name: ARM 64-bit
             runner: ubuntu-22.04
@@ -171,6 +183,7 @@ jobs:
     name: ${{ matrix.platform.name }} Tests
     needs: [get-version, build-linux]
     runs-on: ${{ matrix.platform.runner }}
+    continue-on-error: ${{ matrix.platform.experimental || false }}
     strategy:
       fail-fast: false
       matrix:
@@ -178,10 +191,16 @@ jobs:
           - name: Ubuntu 22
             runner: ubuntu-22.04
             platform-id: ubuntu22
-            
+
           - name: Ubuntu 24
             runner: ubuntu-24.04
             platform-id: ubuntu24
+
+          # See the note on the Ubuntu 26 entry in build-linux: preview image.
+          - name: Ubuntu 26
+            runner: ubuntu-26.04
+            platform-id: ubuntu26
+            experimental: true
 
     steps:
       - name: Checkout
--- a/depends/packages/xproto.mk
+++ b/depends/packages/xproto.mk
@@ -22,5 +22,5 @@ define $(package)_build_cmds
 endef
 
 define $(package)_stage_cmds
-  $(MAKE) DESTDIR=$($(package)_staging_dir) install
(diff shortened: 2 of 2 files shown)


JSanchezFDZ:
Pushed a second commit: the first Ubuntu 26 CI run failed, and it turned out not to be a preview-runner problem but a real bug in the depends build.

```
mkdir: .../include/X11: File exists
make[3]: *** [Makefile:487: install-nodist_xprotoHEADERS] Error 1
```

xproto 7.0.31 installs its headers and its pkgconfig file from separate targets that each create the same directories, and its `install-sh` is not idempotent, so a parallel install can race against itself. Nothing about it is specific to GCC 15 or to 26.04 — it is a timing issue that happens to reproduce reliably on that runner, and in principle it can hit the existing Ubuntu 22/24 and ARM builds too as an intermittent failure.

Fixed by installing that one package with `-j1`, which is exactly what `openssl.mk` already does for its own install step. The step is short so the cost is nil.

JSanchezFDZ:
Added macOS builds (arm64 and x86_64), since you mentioned wanting Mac if it is doable.

**Built natively on macOS runners, not cross-compiled from Ubuntu 26.** Cross-compiling is possible — `depends` has a `darwin` host and the cctools packages — but it needs an *extracted Apple SDK*: `depends/hosts/darwin.mk` points `OSX_SDK` at an Xcode 15.0 (15A240d) SDK that has to be pulled from a `.dmg` under an Apple developer account, and it cannot be redistributed, so CI has nowhere legitimate to fetch it from. A macOS runner already has an SDK, which removes the problem instead of working around it. `depends` supports `arm64-apple-darwin` and `x86_64-apple-darwin` and ships `builders/darwin.mk`, so building on the runner is the supported path.

One thing that would have broken it immediately: both composite actions used `make -j$(nproc)`, and `nproc` is GNU coreutils — it does not exist on macOS. It now falls back to `sysctl -n hw.ncpu`.

Both macOS jobs are marked `experimental` for the moment, the same as Ubuntu 26, so a failure cannot turn the workflow red for the other platforms while this is being shaken out. This is the first time the tree is built for darwin here, so I would rather find out from this PR's own CI run than assert it works. **I cannot verify these locally — I have no macOS machine — so the CI result on this PR is the real test.** If they fail I will iterate here.

They build release binaries only. The debug build and a `.dmg` via `make deploy` are easy to add once the plain build is known to work; I left them out rather than pile untested steps on untested steps.

On the rest of your message: I have **not** removed Ubuntu 22 or ARM 64. "Only need 24, 26, win" could mean either "those are what we ship" or "delete the others", and dropping them would leave 22.04 users (still an LTS) and ARM node operators without binaries. Say the word and it is a two-line change.

JSanchezFDZ:
I have taken the macOS work back out of this PR and will send it separately, so this one is not held up.

The Ubuntu 26 side is finished and was fully green — all five builds and all four test jobs — before macOS was added. macOS turned out to be a bigger job than "add a matrix entry": the darwin host in `depends` has apparently never been built here, and it fails for reasons that have nothing to do with this change. Two of them I have already fixed, and there is at least one more, so it deserves its own PR and its own review rather than sitting on top of a finished one.

This PR is back to exactly what it was when it went green: the Ubuntu 26 build and test jobs, plus the `xproto` install race.

PR #469, opened 2026-08-31
ci: build for macOS on Apple Silicon

JSanchezFDZ:
Split out of #466 so that the finished Ubuntu 26 work is not held up by this.

**Draft: the macOS build does not complete yet.** Opening it anyway because what it has already turned up is worth seeing: the `darwin` host in `depends` appears never to have been built, and it is broken in ways that have nothing to do with CI.

### Approach: native runners, not cross-compilation

You mentioned Mac might have to be cross-compiled from Ubuntu 26. It can be — `depends` has a `darwin` host and the cctools packages — but `depends/hosts/darwin.mk` points `OSX_SDK` at an **Xcode 15.0 (15A240d) SDK that has to be extracted from a .dmg under an Apple developer account** and cannot be redistributed, so CI has nowhere legitimate to fetch it from. A macOS runner already has an SDK. `depends` supports `arm64-apple-darwin` and `x86_64-apple-darwin` and ships `builders/darwin.mk`, so building on the runner is a supported path and avoids the licensing question entirely.

### Fixed so far

**1. `nproc` does not exist on macOS.** Both composite actions ran `make -j$(nproc)`; `nproc` is GNU coreutils. Now falls back to `sysctl -n hw.ncpu`. This would have stopped the build on its first command.

**2. `C_STANDARD` and `CXX_STANDARD` are never defined.** `hosts/darwin.mk` builds its flags as

```make
darwin_CFLAGS=-pipe -std=$(C_STANDARD) ...
darwin_CXXFLAGS=-pipe -std=$(CXX_STANDARD) ...
```

and neither variable is set anywhere in `depends`. They expand to an empty `-std=`, which clang rejects, so every configure test fails:

```
configure: error: C compiler cannot create executables
make: *** [.backtrace_stamp_configured] Error 77
```

Defined them next to the other defaults, with `c++17` to match what `configure.ac` requires. Only the darwin host reads them, so nothing else is affected. **This one means the darwin host cannot ever have built successfully in its current state** — it is not a CI-specific problem.

### Where it stops now

With those two fixed the build gets a long way — boost, Qt and most of `depends` build — and then fails in `qrencode`:

```
qrencode.c:932:9: error: use of undeclared identifier 'VERSION'
        return VERSION;
make[3]: *** [qrencode.lo] Error 1
```

`VERSION` comes from the generated `config.h`, so something about how that package is configured on darwin is leaving it undefined. The same package builds fine for the Linux and ARM hosts.

I have not fixed this one, and I would rather say so than guess: **I have no macOS machine, so I cannot reproduce it outside CI**, and iterating one commit at a time against a runner is a poor way to debug a configure problem. If someone with a Mac runs `make -C depends HOST=arm64-apple-darwin` with these two commits applied, the `config.log` under `depends/work/build/arm64-apple-darwin/qrencode/*/` should say immediately what is missing.

While reading that recipe I also noticed two things in `depends/packages/qrencode.mk` that look unintended, though I have not touched them since I cannot verify the consequences:

- `$(package)_preprocess_cmds` is defined **twice**; the second silently replaces the first.
- both copies end with `cp -f $(BASEDIR)/config.guess $(BASEDIR)/config.sub use` — the destination `use` where every other package in `depends` uses `.`, which would put the updated `config.guess`/`config.sub` somewhere the build does not read them. An out-of-date `config.guess` failing to recognise `arm64-apple-darwin` would be a very plausible cause of exactly this class of failure.

### State of the jobs

Both macOS jobs are marked `experimental`, so they cannot turn the workflow red for anyone else while this is being worked out. They build release binaries only; the debug build and a `.dmg` via `make deploy` are easy to add once the plain build works.


JSanchezFDZ:
Found the qrencode failure, and fixed it — without needing a Mac, because it turns out not to be macOS-specific at all.

**First, a correction to what I wrote above.** I suggested `cp -f $(BASEDIR)/config.guess $(BASEDIR)/config.sub use` looked like a typo for `.`. It is not: `configure.ac` line 19 is `AC_CONFIG_AUX_DIR(use)`, so `use` is libqrencode's auxiliary directory and copying the updated `config.guess`/`config.sub` there is exactly right. I should not have flagged it before checking, and I am sorry for the noise.

**The actual cause.** libqrencode's `autogen.sh` runs the autotools in the wrong order — `autoheader` before `aclocal`:

```sh
autoheader
aclocal $ACLOCAL_OPT
$LIBTOOLIZE --automake --copy
automake --add-missing --copy
autoconf
```

`autoheader` therefore runs before `aclocal.m4` contains the automake macros, and emits a `config.h.in` with `PACKAGE_VERSION` but **no `VERSION`**. `configure` then writes a `config.h` without it, and `qrencode.c:932` — `return VERSION;` — has nothing to return.

**The Linux builds only work by accident.** During `make`, automake's rebuild rules notice `aclocal.m4` is newer than `config.h.in`, re-run `autoheader` (correctly, this time), regenerate `config.h.in` **with** `VERSION`, and re-run `config.status` before `qrencode.c` is compiled. You can see it in the build log as `use/missing autoheader`. Where that mid-build repair does not happen, the latent bug surfaces — which is what darwin did.

**Fix:** run `autoheader` again after `autogen.sh`, so `config.h` is right as soon as `configure` finishes rather than depending on a rebuild rule to patch it up.

Verified on a clean tree, on Linux:

| | `define VERSION` in `config.h` after `configure` |
|---|---|
| before | none at all |
| after | `#define VERSION "4.1.1"` |

and a full `make` still completes and produces `libqrencode.a`, so the Linux and ARM hosts are unaffected.

Also removed a duplicated `$(package)_preprocess_cmds` block — the two were identical, so the second was silently replacing the first.

This is a latent bug in the package recipe rather than anything to do with macOS; darwin just happens to be where it stopped being invisible. Pushed, so CI will now tell us what the next obstacle is.

JSanchezFDZ:
**macOS arm64 builds.** The qrencode fix was the last blocker for it.

I checked the output rather than trusting the green tick, because a 3-minute build next to Ubuntu's 41 minutes looked wrong. It was a legitimate `depends` cache hit, and the artifact is real:

```
raptoreumd:   Mach-O 64-bit arm64 executable   13 MB
raptoreum-qt: Mach-O 64-bit arm64 executable   36 MB
```

So the GUI wallet builds for Apple Silicon, from three commits, none of which needed a Mac to work out.

**The x86_64 job was queued forever because `macos-13` no longer exists.** GitHub retired that image on 4 December 2025, so the job was waiting for a runner that is never coming — it was not slow, it was unschedulable. Switched to `macos-15-intel`, which is the migration label.

Worth knowing for planning: `macos-15-intel` is the **last** x86_64 macOS image GitHub Actions will offer, and it goes away in **August 2027**. After that, Intel macOS builds need a self-hosted runner or have to be dropped. Apple Silicon is the only one with a future here, so if you only want to maintain one Mac target, arm64 is the one to keep.

One thing I noticed but have not changed, since it is pre-existing and affects the Linux jobs too: the `depends` cache key is `depends-${{ github.job }}-<hash of depends/packages>` and does not include the host triplet. Every entry in a matrix shares one key — Ubuntu 22, 24, 26 and ARM all use `depends-build-linux-`, and both macOS jobs use `depends-build-macos-`. It is not incorrect, because `depends` keys its own `built` entries by host, so nothing wrong gets reused; it just means caches carry other architectures' output around. Adding the host to the key would make them smaller and the hit rate more predictable. Happy to do that separately if you want it.

JSanchezFDZ:
Correcting myself on the cache remark above: I said the matrix entries share one `depends` cache key and therefore carry each other's architectures around. The key part is true, the consequence is not.

A cache entry is identified by key **and version**, and the version is derived from the paths being cached. `setup-depends` caches `depends/built` and `depends/${{ inputs.host }}`, so the host is in the path list even though it is absent from the key. Checking the repository's caches shows exactly that — the same key with two different versions:

```
ref=PR469 | ver=4a5d600a6d1f | 136MB | depends-build-linux-...881ea25d
ref=PR469 | ver=6e0531a64b18 | 136MB | depends-build-linux-...881ea25d
```

one for `x86_64-pc-linux-gnu` and one for `aarch64-linux-gnu`. So Ubuntu and ARM already get separate entries and nothing is being mixed. Adding the host to the key would make the entries easier to read, but it would not fix a real problem, and I should not have implied it would.

What *is* shared is Ubuntu 22, 24 and 26, which have the same host triplet and therefore the same key and the same version: all three restore whichever one saved first. In practice that is fine — `depends` output is static libraries — but it does mean the Ubuntu 26 job can be running against dependencies compiled by GCC 11 on Ubuntu 22.

JSanchezFDZ:
**macOS builds for Apple Silicon, and I have dropped the Intel target.** Here is where each one ended up.

### arm64: works, and verified

I did not take the green tick at face value, because a 2-minute build next to Ubuntu's 41 minutes looked wrong. It was a genuine `depends` cache hit, and the artifact is real:

```
raptoreumd:   Mach-O 64-bit arm64 executable   13 MB
raptoreum-qt: Mach-O 64-bit arm64 executable   36 MB
```

Three defects had to be fixed to get there, and none of them are CI problems — they are all reasons the `darwin` host could not have been built by anyone, in any environment:

1. `nproc` does not exist on macOS, so `make -j$(nproc)` failed on the first command.
2. `C_STANDARD` and `CXX_STANDARD` are referenced by `hosts/darwin.mk` but defined nowhere, giving an empty `-std=` that clang rejects.
3. libqrencode's `autogen.sh` runs `autoheader` before `aclocal`, producing a `config.h` with no `VERSION`. Linux only escapes this because a rebuild rule repairs it mid-`make`.

### x86_64: dropped, and why

It took three attempts, each of which found something real:

- `macos-13` no longer exists (retired December 2025), so the job was unschedulable rather than slow.
- `macos-15-intel` ships no Xcode older than 16.0, and Qt 5.15.5 cannot build against it — its bundled libpng includes `<fp.h>`, a Classic Mac OS header absent from the current SDK.
- Cross-building for Intel on the arm64 runner got all the way through `depends` — Qt and everything else — and then failed at link time. `relic`, inside `dashbls`, selects its GMP-backed arithmetic for x86_64, and the link is missing every `__gmpn_*` symbol. GMP *is* built for the host; it is simply not on the link line. arm64 never hits this because relic picks a different backend there.

Fixing that means changing how `dashbls` is linked, which is separate work from getting macOS building at all. I do not think Intel macOS is where that effort belongs: Apple has discontinued the architecture and GitHub removes x86_64 macOS runners in **August 2027**, so anything built there has a two-year shelf life.

So rather than leave a job that is permanently red, the target is removed. The commits that got it as far as the link error are in this branch's history if anyone wants to pick it up.

### Ready for review

The PR now does one thing: it builds and packages Raptoreum for Apple Silicon, and fixes the three `depends` bugs that were in the way. The job is still marked `experimental` so a wobble cannot fail the workflow for other platforms; that is easy to drop once you are happy with it.

npq7721 (commented):
would it possible to get cross build working instead of make it on native mac build so we only need to maintain the build env using for cross build is native mac environment in which it is hard to test without a mac?

JSanchezFDZ:
Yes — it cross-builds now, and you were right to push for it. Pushed: the macOS job runs on `ubuntu-24.04` and produces Mach-O arm64 binaries.

```
raptoreumd      Mach-O 64-bit arm64 executable   19 MB
raptoreum-cli   Mach-O 64-bit arm64 executable  557 KB
raptoreum-qt    Mach-O 64-bit arm64 executable   46 MB
test_raptoreum  Mach-O 64-bit arm64 executable   23 MB
```

all four ad-hoc signed (`LC_CODE_SIGNATURE` present) with `minos 13.0`, built end to end on a Linux box with no Mac involved at any point.

### The one thing that still needs a Mac, and why it cannot be helped

`hosts/darwin.mk` builds against an Apple SDK, and an Apple SDK is not redistributable: CI has nowhere legitimate to download one from. A macOS runner already has Xcode, so there is now a separate `macOS SDK` job that takes the SDK off the runner and hands it to the Linux job as an artifact. **That job compiles nothing** — it copies a directory and uploads it. Everything that has to be maintained, and everything that can break, is in the Linux job.

If you would rather have no macOS runner in the workflow at all, the alternative is for the project to host the SDK tarball itself and inject it through a secret. That works, but PRs from forks cannot read secrets, so the macOS job would fail for every outside contributor. I would rather keep the SDK job.

### What it took: six defects, none of them about CI

The `darwin` host in `depends` has never been built by anyone. What is in the tree is Bitcoin v28's `hosts/darwin.mk` — it arrived in #418, a PR about asset i18n — sitting on top of the toolchain packages from years earlier. v28 expects clang, lld and the llvm binutils on the build machine and ships no `native_cctools` at all; we still built `native_clang` **10.0.1, released in 2020**, plus `native_cctools` and `native_libtapi`. Those three are now removed: nothing asks for a `$(host)-` prefixed toolchain any more.

Then, in the order they showed up:

1. **`-fuse-ld=lld` has to be on the compiler, not just on the linker flags.** `gmp`'s configure links its test program with `$CC` and no `LDFLAGS`, clang falls back to the build machine's GNU ld, and the build stops at `could not find a working compiler` with `/usr/bin/ld: unrecognised emulation mode: llvm` in `config.log`.

2. **`llvm-libtool-darwin` as `LIBTOOL`.** `miniupnpc` and `libnatpmp` archive with Apple's `libtool -static`, which used to arrive as `arm64-apple-darwin-libtool` from cctools.

3. **Qt's `mac-qmake.conf` asks for `${CROSS_COMPILE}ar` and `ranlib`**, where `CROSS_COMPILE` was `$(host)-`. Now `llvm-`, as upstream does. `qtbase` then hands `ranlib` the option `-no_warning_for_no_symbols`, which Apple's has and `llvm-ranlib` does not; upstream comments that line out and so do we.

4. **libpng's `<fp.h>` — and this one matters beyond this PR.** I wrote above that Qt 5.15.5 cannot build against Xcode 16, which is why I dropped the Intel target. That diagnosis was wrong. libpng includes `<fp.h>`, a Classic Mac OS header, whenever `TARGET_OS_MAC` is defined, and **clang predefines `TARGET_OS_MAC=1` for every Apple target**:

   ```
   $ clang --target=arm64-apple-darwin -isysroot <sdk> -nostdlibinc -dM -E /dev/null | grep TARGET_OS_MAC
   #define TARGET_OS_MAC 1
   ```

   A file escapes it only if it happened to include `<math.h>` first, because the inner guard tests `__MATH_H__`. So it is not about Xcode 16 or about macOS; it is a latent bug that surfaces according to the include order of each translation unit. libpng dropped the branch upstream, and this PR does the same. **`macos-15-intel` may well build once this patch is in** — I have not re-tested it, but the reason I gave for abandoning it does not hold.

5. **libtool picked Linux link rules for a darwin host.** It asks the compiler where `ld` is, and `clang -print-prog-name=ld` answers `/usr/bin/ld` even with `-fuse-ld=lld` on the command line, so `with_gnu_ld=yes`. `LD` now travels through `config.site` the way `AR`, `RANLIB` and `NM` already did.

6. **And the one I enjoyed least.** libtool decides whether the linker supports `-single_module` by grepping the linker's stderr for the string `single_module`. lld links the test library successfully — exit 0, dylib produced — and mentions the flag while doing so:

   ```
   ld64.lld: warning: Option `-single_module' is deprecated in ld64:
   ld64.lld: warning: Unnecessary option: this is already the default
   ```

   libtool reads that as a rejection and switches to its two-step recipe, whose first step is `$CC -r -keep_private_externs -nostdlib`. **lld's Mach-O backend does not implement `-r`.** It says so in a warning, links something else entirely, and the build dies on `undefined symbol: strlen`, `strcat`, `free`, `calloc` and the rest of libc. Seeding `lt_cv_apple_cc_single_mod` answers the question directly, which is what `config.site` exists for.

Two more things worth knowing:

- **The ad-hoc signature is no longer suppressed.** `-Wl,-no_adhoc_codesign` sat in the cross-compilation branch of `darwin.mk` only, so it would have applied to exactly the binaries that need a signature most: arm64 macOS refuses to execute a binary carrying no signature at all. It is worth having for deterministic builds; this is not one.

- **`libraptoreumconsensus` is not built for macOS.** It is declared `-no-undefined` and it is not: it references `gArgs`, `GetDataDir`, `BCLog::Logger` and `boost::filesystem`. An ELF shared object may leave those to be resolved at load time, a Mach-O dylib may not, so the link fails. None of the packaged artifacts is that library, so the job passes `--disable-shared`. Deciding what belongs inside that library is a real change and not a CI one, so I have left it alone and flagged it here rather than quietly reshuffling it.

`OSX_SDK` and `OSX_SDK_VERSION` are overridable now, which is how the job points at whatever SDK the runner happens to carry, and `depends/README.md` says what a Linux machine needs to reproduce this by hand: clang 18 or later, `lld`, `llvm`, and `/usr/lib/llvm-18/bin` on `PATH`.

The job is still marked `experimental`, so it cannot fail the workflow for anyone else while you look at it.

JSanchezFDZ:
Green across the board, and I checked the artifact rather than the tick.

```
success  macOS SDK            59s
success  macOS arm64 Build    30m
success  Ubuntu 22 / 24 / 26 Build + Tests
success  ARM 64-bit Build
success  Windows 64-bit Build + Tests
```

Unpacking `raptoreum-macos-arm64-2.0.4.99-SNAPSHOT`:

```
raptoreum-cli    431 KB   Mach-O 64-bit arm64 executable
raptoreumd        12 MB   Mach-O 64-bit arm64 executable
raptoreum-qt      35 MB   Mach-O 64-bit arm64 executable
```

all three carrying `LC_CODE_SIGNATURE`, so they are ad-hoc signed and will actually start on Apple Silicon. Built on `ubuntu-24.04`. The SDK job took 59 seconds, which is the whole remaining cost of having a Mac in the workflow.

**One thing that check caught, now fixed.** The release tarball was 142 MB and had `macos-sdk.tar.gz` inside it: `package-artifacts` collects tarballs with `mv *.tar.gz` from the workspace root, and the SDK artifact was being downloaded into exactly that directory. So the first green run shipped a 94 MB copy of Apple's SDK inside the release artifact — the one outcome this whole arrangement exists to prevent. It now downloads into a subdirectory and is deleted once unpacked into `depends/SDKs`. Pushed as a separate commit; worth a look, since it is the kind of thing a passing build hides.

PR #470, opened 2026-09-11
Functional test suite: make it run, then fix what it found

drmancini:
## Functional test suite: make it run, then fix what it found

The functional suite has not run for some time. Two independent faults
stopped it before a single test executed, so nothing in `test/functional`
was protecting anything.

This branch repairs the suite, ports the tests the fork never took from
upstream, adds coverage for the features Raptoreum has and Dash does not,
and fixes the defects the working suite then exposed.

**149 tests pass, in 398 seconds at `-j12`.** Every commit builds, and
every source fix arrives with the test that catches it.

### Why nothing ran

- `test_runner.py` refused to start. It gated on a config key that
  `configure` has never emitted, so it printed a message about wallet
  support and exited before running anything.
- Every script failed at import. `messages.py` imports a hash module at
  module scope, and the one it named was Dash's X11, not GhostRider. CI
  cloned it from a repository that now returns 404.

The first is a one-line fix. The second needed `contrib/ghostrider-py`,
a small extension that binds the node's own `hash_selection.cpp` rather
than reimplementing it, so the tests cannot drift from consensus.

### What the working suite found

Fifteen source defects, each fixed here with a regression test. The
source diff is small: 18 files, +175/-47.

| Area | Defect |
| --- | --- |
| `rpc/rpcassets.cpp` | Asset RPCs take `cs_main` before `cs_wallet`; the wallet takes them the other way on the scheduler thread. An AB-BA deadlock that stops the node answering. |
| `llmq/quorums.cpp` | Connections to quorums that age out of the window are never closed; the delete-set was read but never emptied. |
| `rpc/misc.cpp` | `echo` declares no arguments and enforces that, so the RPC that exists to echo its arguments rejects all of them. |
| `rpc/misc.cpp` | `deriveaddresses` type-checks the range as a number, so the `[begin,end]` form its own help example uses is refused and a range can only start at zero. |
| `wallet/rpcdump.cpp` | `importprivkey` overwrites an existing label, contradicting the default its help documents. |
| `wallet/fees.cpp` | `-fallbackfee=0` is ignored: the rate is raised to the relay floor and the reason tag overwritten, so the wallet sends at one satoshi per byte instead of refusing. |
| `wallet/rpcwallet.cpp` | `listsinceblock` does not filter conflicted transactions by depth. |
| `wallet/rpcwallet.cpp` | `getaddressinfo` omits the `ischange` field it documents. |
| `rpc/smartnode.cpp` | `smartnode payments` ignores its count argument (half-applied backport). |
| `rpc/server.cpp` | `getrpcinfo` omits the `logpath` field it documents. |
| `logging.cpp` | The category list is inverted, printing only the aliases it means to skip. |
| `init.cpp` | `-asmap` lost its registration in a merge, so the option no longer parsed. |
| `init.cpp` | `-llmq-qvvec-sync` is not validated at startup; a malformed value surfaces much later. |
| `noui.cpp` | Messages print as `ErrorSomething` because the caption carries no separator. |
| `util/system.cpp` | The "specified data directory" error is misspelled, reads the wrong argument name, and eats a quote, so it never names the directory. |

The deadlock deserves a note on evidence. I could not win the race from
a functional test in 390 attempts, so the test committed with it does
**not** claim to catch it. It is proven instead with `-DDEBUG_LOCKORDER`,
which reports the inversion deterministically and is silent once fixed.
Turning that on in CI would be worth doing separately.

### Behaviour pinned but deliberately not changed

Three defects are asserted as they stand, with comments saying so, and
will fail loudly if anyone reconciles them:

- The wallet hands out future outputs that consensus will not let it
  spend, so `sendtoaddress` files a transaction the node has refused.
- `createasset` accepts distribution types that can never exist.
- `createasset` reads `isunique` while its argument list documents
  `is_unique`, so the documented spelling silently yields an ordinary
  divisible asset.

### New coverage

Twenty test files are new. Six cover things with no upstream
equivalent and no coverage at all until now: the asset lifecycle, the
rules the asset RPCs enforce, future transactions, the founder fee, and
the block and wallet query RPCs nothing reached. The other fourteen are
upstream scripts the fork never took.

Eight more sat in the extended list that upstream runs by default, so
their coverage never ran here. They run by default now.

### Reviewing this

The series is ordered so each commit stands alone: infrastructure, then
framework repairs, then per-area test repairs, then source fixes with
their tests. It is large because several test commits only pass because
of a source fix earlier in the series, so splitting it leaves the first
half red.

### Known gaps

- `feature_bip68_sequence.py` and `feature_assumevalid.py` are left as
  they were. The former checks behaviour before a versionbits activation
  Raptoreum does not have, since CSV is unconditionally on.
- `feature_pruning.py` calls `resendwallettransactions`, an RPC this tree
  does not have. Upstream removed both; porting their replacement block
  builder is its own change.


drmancini:
I'll re-submit in more smaller PRs

PR #471, opened 2026-09-13
[1/9] contrib: add raptoreum_hash, a GhostRider module for the test framework

drmancini:
The functional test framework builds and mines blocks in Python. Without a hash
function that matches consensus it cannot produce a header the node will accept, so
every test that mines a block by hand is unrunnable.

This adds `contrib/ghostrider-py`, a small CPython extension that binds Core's own
GhostRider implementation rather than reimplementing it. There is exactly one hash
function in the tree and the tests now use it.

No `src/` change. Nothing outside `contrib/` is touched.

The commit body explains the `PY_SSIZE_T_CLEAN` requirement: the `y#` argument format
needs that macro defined before `Python.h`, Python 3.10 through 3.12 raise
`SystemError` at call time without it, and 3.13 relaxed the rule. A module built and
smoke-tested only on 3.13 therefore looks correct while failing on every version the
rest of the suite can run on.

**Suite at this point:** still cannot run. The runner gates on a config key the build
does not emit; that is fixed in PR 4.

---

Part 1 of 9, replacing #470. First in the series. Targets `develop`; until the earlier parts merge, this diff also shows their commits, so review the commits rather than the combined diff.


JSanchezFDZ (approved):
Revisión de `2998729e1ddb780ecd25078dd6374833b7d335aa`, los cuatro archivos y el recorrido hasta `HashGR`/CryptoNight. **Aprobable para las cabeceras de 80 bytes que utiliza el framework; no encuentro problemas P1/P2.**

Comprobé la separación entre identidad SHA256d y PoW GhostRider, el orden de los bytes del hash anterior, la vida del argumento al liberar el GIL y la independencia de los contextos de cálculo.

**Pruebas ejecutadas**

- Construcción e instalación reales mediante `python3 -m pip install --no-build-isolation --no-deps --target ... contrib/ghostrider-py`, con Python 3.12, GCC/G++ 13 y Boost 1.83: rueda creada, importada y ejecutada correctamente.
- Rechazo de tipos y longitudes inválidos; contrato mínimo de 36 bytes y salida de 32 bytes comprobados.
- Reconstrucción de los génesis mainnet/testnet/regtest: identificadores SHA256d correctos y PoW bajo sus targets. 63 resultados concurrentes en cuatro hilos coinciden con el cálculo secuencial.
- Comprobación sintáctica C++14 correcta; no considero la ausencia de un flag C++17 explícito un error demostrado. CI del SHA revisado correcta.

No ejecuté la suite funcional completa ni una matriz de instalación en Windows/macOS; estas pruebas validan el módulo en Linux x86-64.

**P3, robustez no bloqueante:** [ghostridermodule.cpp:36–47](https://github.com/Raptor3um/raptoreum/blob/2998729e1ddb780ecd25078dd6374833b7d335aa/contrib/ghostrider-py/ghostridermodule.cpp#L36) acepta cualquier longitud superior al mínimo como `Py_ssize_t`, pero `HashGR` la estrecha a `int`. Una entrada excesiva puede truncarse silenciosamente y producir un hash de menos bytes de los solicitados; lo confirmé con la extensión compilada. No afecta al uso normal de 80 bytes ni demuestra un problema de consenso. Rechazar longitudes superiores a `INT_MAX`, o exigir exactamente 80 bytes si ese es el contrato elegido, y añadir una prueba de rechazo del límite.

Documentaría también los requisitos de compilador, cabeceras Python y Boost para la instalación desde fuentes. Una prueba pequeña con vectores fijos y concurrencia protegería el binding sin depender de toda la suite funcional.

drmancini:
Thanks for the catch. Fixed in 5e6e81eff.

While implementing the fix I found the issue is more severe than "P3, non-blocking" — reproduced it directly: calling the original binding with a real buffer just above `INT_MAX` bytes doesn't just truncate silently, it **segfaults the whole process**. The narrowed `int` wraps negative, and every hash primitive `HashGR`'s first round can dispatch to (`sph_*512`) takes a `size_t` length — so the negative value gets reinterpreted as an ~18-exabyte unsigned length, causing an out-of-bounds read.

Fix: reject any length above `INT_MAX` before `HashGR` is ever called, at the boundary this binding owns.

Also added, since there was no test coverage for this module at all before:
- `test_binding.py` — the documented 36-byte minimum, the new `INT_MAX` rejection (verified against the actual regression: reverting the fix crashes the whole test run rather than failing one test cleanly), and known-answer vectors from three real mainnet blocks (heights 1, 500000, 1000000), each independently verified against its own block's real PoW target. (The previous algorithm-selection test only proved two different headers hash differently, which is trivially true regardless of whether prev-hash extraction is correct — replaced it with the KAT vectors instead.)
- Documented the compiler / Python-header / Boost / setuptools build prerequisites, and the read-only-buffer requirement, in the README.

PR #472, opened 2026-09-13
[2/9] test: remove tests that no longer describe this node

drmancini:
Three kinds of dead weight, one commit each:

- Seven stale copies of tests that were renamed upstream. Both names exist in the
  tree; the old copies were never updated and test nothing current.
- The governance test. Smartnodes do not run governance on Raptoreum; it is a Dash
  relic.
- The block reward reallocation test, which encodes Dash's reallocation schedule
  rather than this chain's.

Deletions only, apart from one line in `test_runner.py`.

**Suite at this point:** still cannot run, same gate as above.

---

Part 2 of 9, replacing #470. Builds on #471. Targets `develop`; until the earlier parts merge, this diff also shows their commits, so review the commits rather than the combined diff.


JSanchezFDZ (changes requested):
Reviewed `abd220938bc494d542b6d6b47e2320cd94800225`: complete #471 → #472 increment, including all seven old/new test pairs, the removed reward-reallocation test, and the runner. **Requesting one inventory correction; the other deletions are justified.**

**P2 — Explicitly register the governance exclusion so `--ci` does not abort.**

[test_runner.py:174](https://github.com/Raptor3um/raptoreum/blob/abd220938bc494d542b6d6b47e2320cd94800225/test/functional/test_runner.py#L174) comments out `feature_governance_objects.py` but leaves the file in the enumerated directory. `check_script_list()` compares all Python files against `ALL_SCRIPTS + NON_SCRIPTS`; a comment is not an exclusion. With `--ci`, the unregistered file causes exit 1 before tests run. The integration-test wrapper uses that option.

I executed the actual checker from #471, #472, and #479 with each commit's exact `git ls-tree` inventory supplied as directory contents. After excluding unregistered files already present in #471, the base check passes and #472/#479 fail solely for governance. The full base inventory also had pre-existing problems; I am **not** claiming the complete runner passed before this PR. This change introduces an additional independent failure and it is not repaired later in the series.

Add an explicit documented exclusion or move the deliberately disabled file outside the enumerated directory. There is no need to re-enable governance or weaken the general completeness check to fix this.

**Coverage reviewed**

The seven old copies were already absent from the runner; their replacements remain present and registered. I compared their cases, including renamed CoinJoin RPCs, configuration, notifications, comments, limited-network service, and quorum connections. Removed deprecated RPC cases refer to contracts already unavailable in the node. The removed Dash reward-reallocation test asserts a deployment/payment schedule this consensus does not implement; current smartnode-payment checks remain.

Governance exclusion is a coverage decision, not proof that governance RPCs have disappeared. Document it accordingly.

**Checks and limits**

Seven successor/inventory checks, 21 syntax checks, the real inventory-checker executions above, and `git diff --check` passed with the expected failures explicitly asserted. No remaining references to the deleted filenames were found. I did not compile/run the daemon or execute a functional suite. Existing green CI checks run the unit-test binary and do not validate this Python runner path. Add a small runner check distinguishing registered, deliberately excluded, and accidentally forgotten files.

drmancini:
Fixed in b4f9f47c8.

Added a `DISABLED_SCRIPTS` list — separate from `NON_SCRIPTS`, which is specifically for files that aren't test scripts at all — as a documented home for real test scripts deliberately excluded from the registered lists, and wired it into `check_script_list()`. `feature_governance_objects.py` is registered there now. Didn't touch the general completeness check or re-enable governance, per your note.

Reproduced the exact failure directly before fixing it: `check_script_list(fail_on_warn=True)` exits 1 on the pre-fix tree, citing `feature_governance_objects.py` among the missing scripts; confirmed it no longer appears after the fix.

One thing worth flagging: reproducing this also surfaced that the same check independently fails on three other files — `rpc_dumptxoutset.py`, `rpc_getblockfilter.py`, `p2p_blockfilters.py`. I confirmed these are a pre-existing baseline problem, not something this PR introduces — the commit immediately before #471 already fails the same way for all three, so I left them out of scope here.

PR #473, opened 2026-09-13
[3/9] regtest: register test-only quorum types

drmancini:
Two problems stop a regtest chain from ever forming a quorum.

The test quorum types are not registered on regtest, so nothing can sign. And
`UpdateLLMQParams` rescales quorum parameters to the smartnode count on every block,
which overwrites whatever size a test chose through `-llmqtestparams` and re-registers
production-sized types that a handful of local nodes can never fill.

This registers the test types on regtest and makes `UpdateLLMQParams` return early
there. Scaling to smartnode count remains a mainnet and testnet concern, untouched.

This is the only consensus-adjacent code in the series, and it is inside a regtest
branch. Mainnet and testnet behaviour is unchanged.

**Suite at this point:** unit tests pass; the functional suite still cannot run.

---

Part 3 of 9, replacing #470. Builds on #472. Targets `develop`; until the earlier parts merge, this diff also shows their commits, so review the commits rather than the combined diff.


JSanchezFDZ (changes requested):
Reviewed `edae3e43e4203d50c5443d45ea47132130a8c4bc`: complete #472 → #473 increment (four files), quorum consumers, parser, and isolation from other networks. **Requesting one regtest retention fix.**

**P2 — Initialize `keepOldKeys` for the new `llmq_test` quorum.**

The initializer at [quorums_parameters.h:428–429](https://github.com/Raptor3um/raptoreum/blob/edae3e43e4203d50c5443d45ea47132130a8c4bc/src/llmq/quorums_parameters.h#L428) omits `keepOldKeys`, leaving it zero. Therefore `max_store_depth()` is zero, unlike `llmq_test_v17` and the previous small quorum, which retain four intervals.

When the hourly smartnode maintenance runs, `CleanupOldContributions()` deletes records whose quorum base is older than the current tip, even if the quorum is still active. For a quorum at height 300 and tip 310, this deletes verification, secret, and encrypted contributions; subsequent encrypted-contribution retrieval can return `ENCRYPTED_CONTRIBUTIONS_MISSING`, breaking key recovery from that node. Timing during an unfinished DKG can also remove contributions before aggregation.

I compiled the actual declarations/initializers in an isolated C++ check: `keepOldKeys=0`, and the cleanup predicate for 300 → 310 evaluates to deletion for the new quorum but not for the old or corrected configuration. The database/recovery consequence is established by tracing the actual cleanup and QGETDATA paths; I did not reproduce it in a complete node. Add `.keepOldKeys = 4`, matching `llmq_test_v17`, and test cleanup/recovery while the quorum remains active. All four changed files remain identical in #479.

**Other checks and limits**

Mainnet/testnet/devnet constructors and the non-regtest remainder of `UpdateLLMQParams` are unchanged; I checked both callers and registration/selection of the test types. DKG windows and the override's size/minimum/threshold fields are consistent. `git diff --check` passed. Parser boundary cases were reviewed statically; an attempted separate parser compilation could not proceed because that isolated environment lacked the required Boost header. No complete build, node, or DKG scenario was executed. Add invalid-format/range parser tests and verify both test types' retention settings explicitly.

drmancini:
Fixed both. `.keepOldKeys = 4` added to `llmq_test` (`67f2436e`), matching `llmq_test_v17`.

Verified directly: reverted the fix, rebuilt, and confirmed the failure mode you traced -- `max_store_depth()` came back `0` instead of `120`. Restored the fix, rebuilt, confirmed `120` on both types, and ran the full unit suite clean.

Also added `src/test/chainparams_tests.cpp` per your note on parser coverage and explicit retention checks:
- `regtest_llmq_types_share_retention_window` -- pins `keepOldKeys`/`max_store_depth()` explicitly equal for both test types, as a regression guard for this exact bug.
- `llmqtestparams_applies_valid_size_and_threshold` -- valid `size:threshold` applies to both types.
- Six rejection cases against the real `-llmqtestparams` parser: missing colon, extra colon, non-numeric values, threshold>size, zero size, zero threshold -- each confirmed to throw `std::runtime_error` from the actual `CRegTestParams::UpdateLLMQTestParametersFromArgs`, not a mock.

All four changed files remain identical to what you noted persists through #479.

PR #476, opened 2026-09-13
[6/9] rpc, llmq and wallet fixes found by the suite

drmancini:
Six defects, each with the test that catches it:

- `listsinceblock` did not filter conflicted transactions by depth.
- `smartnode payments` ignored its `count` argument.
- `logging` listed category aliases rather than the real categories.
- `getrpcinfo` did not return the `logpath` field it documents.
- Connections to quorums that fall outside the active window were never closed.
- The asset RPCs took `cs_main` before `cs_wallet`, the opposite order to the rest of
  the wallet, which is a lock inversion.

**Suite at this point:** 117 passed, 5 failed, 3 skipped: the same
`rpc_deriveaddresses`, `rpc_named_arguments` and `wallet_createwallet` entries, all
resolved by PR 7 and PR 8. The three single-stage flakes above do not recur.

---

Part 6 of 9, replacing #470. Builds on #475. Targets `develop`; until the earlier parts merge, this diff also shows their commits, so review the commits rather than the combined diff.


JSanchezFDZ (changes requested):
Reviewed `26733a367314e4414334a9259c4eb09c18801444`, focusing on the complete #475 → #476 increment (14 files). **Requesting changes for three confirmed issues.** All three also remain in #479.

**1. P1 — Preserve synchronization of the global asset cache.**

At [rpcassets.cpp:656](https://github.com/Raptor3um/raptoreum/blob/26733a367314e4414334a9259c4eb09c18801444/src/rpc/rpcassets.cpp#L656) and [832](https://github.com/Raptor3um/raptoreum/blob/26733a367314e4414334a9259c4eb09c18801444/src/rpc/rpcassets.cpp#L832), `sendasset` and `listassetsbalance` lose `cs_main` while retaining calls to `passetsCache->GetAssetMetaData`. `cs_wallet` protects one wallet, not this global cache. `GetAssetMetaData` reads/copies metadata and may insert entries; it has no internal mutex. Block connection/disconnection updates the same map through `CAssetsCache::Flush` under `cs_main`.

A balance query concurrent with a mint/update can copy metadata while another thread modifies it: a C++ data race and undefined behavior. I traced the dispatcher and called helpers; none retains `cs_main` around the subsequent cache access. Both RPCs held it before this change. Preserve `cs_wallet` → `cs_main`, as the other RPCs corrected here do, or take an explicitly protected metadata snapshot. This is a complete static trace; I did not run ThreadSanitizer against the daemon.

**2. P2 — Validate the negative integer boundary when enabling `count`.**

The correction at [smartnode.cpp:400](https://github.com/Raptor3um/raptoreum/blob/26733a367314e4414334a9259c4eb09c18801444/src/rpc/smartnode.cpp#L400) makes the argument reach `nCount`. The parser accepts the minimum `int64_t`, but `std::abs(nCount)` at line 405 overflows before conversion to `uint64_t`. The old guard left `nCount=1`, so this argument could not previously expose that boundary. Compiling the exact expression with UBSan confirmed an unrepresentable negation. This was an isolated arithmetic check, not a reproduced RPC crash. Reject that boundary or compute the unsigned magnitude safely; also cover zero and iteration down to genesis.

**3. P2 — The new `mallocinfo` test fails when that capability is unavailable.**

[rpc_misc.py:48–50](https://github.com/Raptor3um/raptoreum/blob/26733a367314e4414334a9259c4eb09c18801444/test/functional/rpc_misc.py#L48) expects `mallocinfo mode not available`, whereas the branch without `HAVE_MALLOC_INFO` returns `mallocinfo is only available when compiled with glibc 2.10+`. The actual `assert_raises_rpc_error` helper rejects that response; the control using the server's literal passes. Correct the expectation so supported builds without this function, including macOS/Windows, do not fail the test.

**Validation and coverage**

Read the entire increment, callers, and locking paths. Syntax checks on all seven changed Python files and `git diff --check` passed. Executed checks of invariants/models for LLMQ, logging, and depth, plus the UBSan and RPC-helper checks above. I did not rebuild the daemon or run the complete functional suite; existing green checks do not establish that these races/boundaries are safe.

I found no new regression in LLMQ connection retention, logging-category exposure, or `listsinceblock` conflict filtering. Add concurrent asset query/send versus chain-update coverage and boundary cases for `smartnode payments`. Clarify the documentation: `target_confirmations` still determines `lastblock`, rather than filtering `listsinceblock` transactions. Update the `ListAssets` comment to reflect the newly added assertion.

drmancini:
All three fixed.

**P1**: restored `cs_main` after `cs_wallet` in `sendasset`/`listassetsbalance`, matching the order every other wallet RPC already uses to reach it. `listunspentassets` (touched by the same earlier commit) never accesses `passetsCache`, confirmed by reading it in full -- needs no change. Ran the full `feature_assets.py` lifecycle test end-to-end against a rebuilt daemon -- passes clean, no deadlock.

**P2 (count)**: rejected the `INT64_MIN` boundary explicitly. Testing down to genesis (per your suggestion) surfaced a second, more serious bug: querying payments at or through genesis segfaults the daemon -- `pindex->pprev` is null there and dereferenced unconditionally for the subsidy, plus a second unguarded path through `GetBlockTxOuts`'s own `ChainActive()[nHeight-1]` lookup. Fixed by skipping genesis in the RPC's own loop. Confirmed with a real RED/GREEN cycle -- reverted, rebuilt, watched it segfault under the new test, restored, confirmed clean against a real 10-smartnode network. Added boundary coverage (count=0, `INT64_MIN`, negative count past genesis) to `rpc_masternode.py`.

**P2 (mallocinfo)**: corrected the expected string. Confirmed by temporarily undefining `HAVE_MALLOC_INFO` and rebuilding to actually exercise the error path, not just compare strings.

**Also**: `CWallet::ListAssets()` carried a stale upstream TODO describing a lock state an earlier commit in this stack already fixed -- replaced with an accurate comment. Added a code comment at `listsinceblock`'s filter distinguishing `depth` (from `blockhash`) from `target_confirmations` (which only affects `lastblockhash`, as the help text already correctly says).

**New, separate finding -- not fixed here**: `protx list valid|registered` with `detailed=true` takes `cs_main` then reaches `cs_wallet` via `CheckWalletOwnsKey` -- the opposite order from every other wallet RPC, a real deadlock risk. Pre-existing, untouched by this PR's diff, flagging for a separate look.

PR #477, opened 2026-09-13
[7/9] test: cover what nothing reached

drmancini:
New tests, no behaviour changes:

- Asset RPCs, what they accept and what they refuse.
- Future transactions.
- The founder fee.
- The query RPCs no test reached at all.
- Upstream tests the fork never took.
- `wallet_createwallet.py`, which was simply missing.

Three repairs to existing tests are included because they are coverage gaps rather
than failures: `feature_addressindex` picked an output at random, several tests
half-covered the RPC they named, and `create_block` now takes the node so hand-built
blocks are valid rather than accidentally so.

One of those repairs is worth calling out. `rpc_scantxoutset` pays its three scan
targets to the wallet's own addresses, so the thirteen sends that follow could select
and spend those very outputs, leaving the totals short by some subset of 0.001, 0.002
and 0.004. Measured at one failure in twenty-four runs. The outputs are now locked as
they are created, and their presence in the UTXO set is asserted before the wallet is
deleted, so a recurrence names its cause rather than surfacing as a wrong total forty
lines later. Ninety-six runs clean afterwards.

**Suite at this point:** 132 passed, 5 failed, 3 skipped. `rpc_deriveaddresses`
(twice) and `rpc_named_arguments` wait on PR 8. `wallet_address_queries` waits on the
`importelectrumwallet` fix in PR 8. `feature_llmq_is_cl_conflicts` appears at this
stage only.

---

Part 7 of 9, replacing #470. Builds on #476. Targets `develop`; until the earlier parts merge, this diff also shows their commits, so review the commits rather than the combined diff.


JSanchezFDZ (changes requested):
Reviewed `9ee7c8518e09ebeeff66e369b2e4dcf999bc2201`: the complete #476 → #477 increment (27 files). **Requesting changes for two sources of test flakiness.** Two additional tests can pass without checking the property they announce. All four issues remain in #479.

**1. P2 — Registering a peer after starting the thread without synchronization.**

[p2p_leak.py:125–126](https://github.com/Raptor3um/raptoreum/blob/9ee7c8518e09ebeeff66e369b2e4dcf999bc2201/test/functional/p2p_leak.py#L125) adds a connection while `NetworkThread` is already running. That thread iterates `mininode_socket_map.items()` without taking the lock, and `peer_connect` registers the connection before initialization is complete. The previous test created peers before starting the thread. Controlled scheduling of the actual method reproduced `RuntimeError: dictionary changed size during iteration`. Create connections before starting the thread, or synchronize registration and iteration consistently; locking only the writer is insufficient.

**2. P2 — Checking the log before the remote action completes.**

[rpc_setban.py:29–31](https://github.com/Raptor3um/raptoreum/blob/9ee7c8518e09ebeeff66e369b2e4dcf999bc2201/test/functional/rpc_setban.py#L29) exits `assert_debug_log` when `addnode` returns, which does not ensure the receiving node has processed/logged the rejection. This framework's context manager reads the file once without waiting. Running the real context with a delayed write confirmed failure before the correct message arrives. Wait for rejection/disconnection or poll for the message with a deadline inside the context. I did not measure daemon failure frequency.

**3. P3 — Wallet maturation can go unchecked.**

At [feature_futures.py:213–215](https://github.com/Raptor3um/raptoreum/blob/9ee7c8518e09ebeeff66e369b2e4dcf999bc2201/test/functional/feature_futures.py#L213), the final assertion only runs if `listunspent` contains the entry. The earlier rejected spend remains in the wallet, and `IsSpent` hides its input until abandonment. The actual method passes with that state without ever observing `futureSpendable=True`. Abandon the rejected spend or use another untouched future, require the entry to exist, and check the flag unconditionally.

**4. P3 — The privacy test can pass without exercising rejection.**

[p2p_leak_tx.py:47–55](https://github.com/Raptor3um/raptoreum/blob/9ee7c8518e09ebeeff66e369b2e4dcf999bc2201/test/functional/p2p_leak_tx.py#L47) succeeds if all 100 attempts are announced before `getdata`; it never has to observe `notfound`. I confirmed this by running the actual `run_test()` with a controlled peer. Require at least one attempt meeting the intended precondition, and fail explicitly if all retries are exhausted.

**Validation performed**

Reviewed the complete delta, related wallet/network flows, and survival through #479. AST checks for all 27 files and `git diff --check` passed. The four checks above execute Python methods extracted from this HEAD with controlled responses/scheduling: they verify those flows, not the full suite or real-world flake frequency. I did not run the daemon for this SHA.

The #478 dependency for file-extension handling in `wallet_address_queries` is declared and resolved there; it is not reported as a hidden final-series defect. Document that asset tests mainly exercise RPC/mempool barriers rather than every consensus restriction. In `wallet_createwallet`, the section announcing encryption of a keyless wallet never calls `encryptwallet`: remove the repetition or adjust the case and its description.

drmancini:
Both flakiness sources and both weak assertions fixed.

**P2 (`p2p_leak.py`)**: moved the 4th connection to join the other three before `network_thread_start()`, and added a `wait_until` on its handshake completing before `generate(1)` runs -- the assertion was still timing-ordered on my first pass (only coincidentally reliable), caught during a follow-up review. Confirmed `NetworkThread.run()`'s unlocked dict iteration is the real hazard by reading it directly; couldn't trigger the race at natural timing (only under artificial slowdown), so I'm not claiming a live reproduction of the exact `RuntimeError`, just the mechanism.

**P2 (`rpc_setban.py`)**: added a polling variant of `assert_debug_log` (same shape, `wait_until` instead of a single read). Ran 5x against a live 2-node setup.

**P3 (`feature_futures.py`)**: added `abandontransaction` before the maturity check and made the assertion unconditional. Verified with a real RED/GREEN cycle against a live node -- removing the fix reproduces a genuine `KeyError` on the exact locked output.

**P3 (`p2p_leak_tx.py`)**: added an explicit failure if the retry loop exhausts without ever observing `notfound`. Confirmed live that the intended branch is actually the one exercised, not just the guard never firing.

**Minor**: `wallet_createwallet.py`'s dead encryption-test section actually calls `encryptwallet` now -- confirmed live that this fork accepts encrypting a keyless wallet and it changes nothing checked. `feature_assets_rules.py`'s docstring now states it covers RPC/mempool barriers, not full consensus validation.

All six files run against a live daemon, not just reasoned through statically.

PR #478, opened 2026-09-13
[8/9] wallet and rpc fixes found by the new coverage

drmancini:
- `echo` could not echo. Its `RPCHelpMan` declares an empty argument list and the body
  validated against it, so every call with an argument threw the help text, while the
  dispatch table registers `arg0` through `arg9`. The registration and the help
  contradicted each other. This restores upstream's body. The assertion PR 5 commented
  out is restored in the same commit.
- `deriveaddresses` did not accept a `[begin,end]` range.
- `importprivkey` discarded an existing label.
- `-fallbackfee=0` paid the floor instead of being honoured.
- `importelectrumwallet` read the extension with `find_last_of(".")` over the whole
  argument, so a path under a directory called `2.0.3` or `.raptoreum` supplied one and
  a file with no extension at all was reported as having the wrong one. It now asks
  `fs::path` for the extension, as the same function already does forty lines further
  down. This is the only site in the tree that reads an extension this way. The test
  creates the dotted directory itself rather than relying on the build path, so the
  case fails on the old code everywhere rather than only where the tree happens to sit
  under a dot.

**Suite at this point:** 139 passed, 0 failed, 3 skipped. With `pyzmq` installed,
142 passed, 0 failed, nothing skipped.

---

Part 8 of 9, replacing #470. Builds on #477. Targets `develop`; until the earlier parts merge, this diff also shows their commits, so review the commits rather than the combined diff.


JSanchezFDZ (changes requested):
Reviewed `93bee90709e67818859901f975168f6e3cd3e9a3`: the complete #477 → #478 increment (11 files), affected RPC/wallet callers, and added tests. **Requesting a descriptor-range boundary fix.**

**P2 — Widen the loop counter for the newly accepted array ranges.**

The type-check change at [misc.cpp:341](https://github.com/Raptor3um/raptoreum/blob/93bee90709e67818859901f975168f6e3cd3e9a3/src/rpc/misc.cpp#L341) permits array-form ranges. `ParseDescriptorRange` accepts endpoints up to `INT_MAX` and bounds the number of requested entries, but the [derivation loop](https://github.com/Raptor3um/raptoreum/blob/93bee90709e67818859901f975168f6e3cd3e9a3/src/rpc/misc.cpp#L370) uses `int`. After processing a short valid range ending at the maximum accepted index, its final increment cannot be represented by that signed type. This is undefined behavior rather than a defined successful return.

The loop itself predates this PR, but the old numeric-only check rejected array ranges. A scalar maximum implied a range starting at zero and failed the million-entry guard. The new representation therefore exposes a previously unreachable boundary. Use `int64_t` for the loop counter, matching the range variables, while retaining the existing index bounds. Add a regression for the maximum endpoint and adjacent ordinary ranges. This finding is a source-level boundary proof; I did not execute a crash-inducing RPC or claim a reproduced daemon crash. The same code remains in #479.

**Other paths checked**

- The `echo` change permits its documented arguments; `CHECK_NONFATAL` produces a catchable RPC error. The restored CLI/RPC assertions match that contract.
- Returning a zero fallback fee before minimum-fee adjustments preserves `FeeReason::FALLBACK`, allowing `CreateTransaction` to report disabled fallback correctly. I traced both `GetMinimumFeeRate` callers and wallet initialization; explicit fee-selection paths remain separate.
- `importprivkey` preserves an omitted existing label and permits explicit replacement. Extend its tests to assert ownership after import, explicit empty-label replacement, and handling of a previously labelled send address.
- `importelectrumwallet` now derives the extension from the filename; a dot in the parent directory no longer supplies one. The new regression covers this case before file parsing.

**Validation and limits**

Read the entire delta and relevant callers/error handling. All seven changed Python files pass syntax checks; `git diff --check` passes. A read-only check verifies the actual parser/loop contracts and evaluates the endpoint arithmetic with unbounded Python integers; it does not execute the unsafe C++ loop. The relevant sources were compared with #479.

No daemon, full build, functional suite, transaction, or runtime crash test was executed. Existing green unit-test CI does not cover this missing functional boundary. Optional documentation cleanup: `wallet_fallbackfee` still describes replace-by-fee although its assertions exercise fee-estimation fallback.

drmancini:
Fixed, and found two sibling instances.

**P2**: widened the loop counter to `int64_t` in `deriveaddresses`, narrowing back to `int` only at `Expand()`'s call site (already bounded to fit by the same range check). Found the identical pattern in `scantxoutset` and `importmulti` over the same `ParseDescriptorRange` result -- both already accepted array-form ranges starting above zero before this PR, so the boundary was reachable there too. Fixed both the same way.

Added the maximum-endpoint regression plus an adjacent ordinary range, verified live: all three RPCs return correct addresses on `[2147483645, 2147483647]` without hanging. A bounded standalone reproduction (using `int8_t` in place of `int`, to avoid a multi-billion-iteration wait) confirms the actual failure shape once it fires: wraparound that keeps the loop running far past its intended bound, not an immediate crash. One correction to my own thinking here -- I don't have a reproduction of this on the shipped `-O2` release build specifically; the UB is real regardless of what any one build's optimizer does with it, but I want to be precise that I'm not claiming a live crash on the actual binary you'd ship.

**Test coverage**: extended `wallet_import_with_label.py` to assert ownership after `importprivkey` (previously only labels were re-checked, never that the address actually became spendable), plus a case for a previously-labelled send-purpose address surviving an unlabeled import, and explicit-empty-label clearing -- all verified against a live node first.

PR #479, opened 2026-09-13
[9/9] test: run by default what upstream runs by default

drmancini:
Aligns `test_runner.py`'s default set with upstream's, and removes the one genuine
duplicate entry (`p2p_unrequested_blocks.py`). Entries that look duplicated are the
same script under different flags, `--usecli`, `--ipv4` and `--spork21`, and are kept.

**Suite at this point:** 146 passed, 0 failed, 3 skipped.

With `pyzmq` installed the suite is 149 tests with nothing skipped. A clean run is
147 passed and 2 failed, both of them intermittent and neither introduced by this
series. They are described below.

---

Part 9 of 9, replacing #470. Builds on #478. Targets `develop`; until the earlier parts merge, this diff also shows their commits, so review the commits rather than the combined diff.


JSanchezFDZ (commented):
Reviewed `ab152c0399be9f95f2301fc0e30f152faf83995f`, specifically the #478 → #479 increment. **No new P1/P2 defect found in the runner-list change. Integration remains conditional on fixing the preceding reviews and validating the combined functional suite.**

Exactly eight invocations move into the default list. Removing the identical `p2p_unrequested_blocks.py` duplicate makes the default count 142 → 149, while extended-only entries change 15 → 7. The set of distinct invocations and all flag variants are preserved; no extended test is silently lost. AST comparison confirms executable code outside the two lists is unchanged.

**Validation performed**

- Compared the complete delta, registered paths, syntax, selection/exclusion code, promoted scripts, and CI wrappers. All registered paths exist and pass the runner's prefix validation.
- Executed the actual `TestHandler` with process stubs: all 149 default invocations preserve flags and receive distinct temporary paths/port seeds. This checks command construction; it deliberately launches no nodes or functional tests.
- Inspected the exact-head unit-test CI artifact: 363 tests, zero failures/errors/skips. The [workflow](https://github.com/Raptor3um/raptoreum/blob/ab152c0399be9f95f2301fc0e30f152faf83995f/.github/actions/run-tests/action.yml#L19) runs `test_raptoreum`, not this Python suite. Its green status is not functional-suite evidence.

**Requirements for a functional CI gate, inherited from earlier code**

1. The completeness check rejects three unregistered files: `p2p_blockfilters.py`, `rpc_getblockfilter.py`, and `feature_governance_objects.py`. The governance exclusion is addressed in the review on #472. Register deliberate exclusions explicitly without weakening the general check.
2. After that gate, [run_tests():416](https://github.com/Raptor3um/raptoreum/blob/ab152c0399be9f95f2301fc0e30f152faf83995f/test/functional/test_runner.py#L416) references undefined `TRAVIS_TIMEOUT_DURATION` when `runs_ci=True`. Calling the real function with external processes stubbed reproduces `NameError` in both #478 and #479 before a test launches. Define the intended timeout or remove the obsolete dependency before enabling that command as a gate.

Neither obstacle is introduced by these eight promotions. Keep findings attached to their originating changes, including the outstanding reviews on #473–#478. Record a reproducible final-suite run with exact commit, command, platform, dependencies, pass/fail/skip counts, and failure logs. Treat optional skips and intermittent failures explicitly; the author's full-suite count was not reproduced here.

**Non-blocking scheduling/documentation:** `mining_getblocktemplate_longpoll.py` is placed under “less than 30s” although its own comment says approximately 70 seconds in the best case. Update the duration group using measured runs and schedule longer tests earlier if justified. No runtime/performance regression was measured. I did not execute the daemon, full functional suite, or a cross-platform runtime matrix.

drmancini:
Ran the full functional suite twice against this branch (commit `58f875b88`) to confirm the `--ci` fix and get real, reproducible numbers rather than deferring this to "next steps":

**Platform:** Linux 6.8.0-137-generic x86_64, g++ 13.3.0, Python 3.11.9

| Run | Command | Result | Wall-clock | Accumulated |
|---|---|---|---|---|
| 1 | `test_runner.py -j 12 --tmpdir=... --combinedlogslen=4000` | 149/149 passed | 423s | 4673s |
| 2 | `test_runner.py -j 12 --ci --tmpdir=... --combinedlogslen=4000` | 149/149 passed | 404s | 4599s |

0 failed, 0 skipped in both runs, `--ci` mode included (confirming `TRAVIS_TIMEOUT_DURATION` and the script-completeness gate both work as restored).

PR #480, opened 2026-09-16
net: don't busy-loop on peers the receive path will not drain

drmancini:
## What is wrong

`CConnman::SocketHandler()` decides whether to skip waiting for socket events using a
different test from the one that decides which nodes are actually read.

The skip-wait test is `!mapReceivableNodes.empty()` (`net.cpp:1623`). The receive path, forty
lines later, drains a node only when **all three** of these hold (`net.cpp:1718`):

```cpp
!it->second->fPauseRecv && it->second->nSendMsgSize == 0 && !it->second->fDisconnect
```

A node satisfying the first but not the second stays in the receivable set with its readable
flag set. The loop skips its wait, polls with a zero timeout, finds nothing it is willing to
do, and immediately repeats. Nothing notices it is not progressing, and the state persists
until the peer disconnects: the socket thread pinned at 100% of a core, the mempool and the
byte counters static.

Two things make a node undrainable:

- **pause** — the peer's queued message bytes exceed `-maxreceivebuffer`, which at the 5 MB
  default takes a large backlog.
- **send queue** — the node has bytes queued to send to that peer. This is not gated on the
  receive buffer and can arise in ordinary operation.

The second is why this is worth fixing rather than tuning around. A remote peer can reach
and hold either state, so it is remotely triggerable. While the thread spins it still
services other sockets each iteration, so the cross-peer effect is through CPU and
`cs_vNodes` contention rather than blocked I/O — but one connection can pin the single socket
thread, and a node that accepts inbound connections can be held this way cheaply.

## How to see it

`contrib/devtools/repro-socket-busyloop.py` reproduces both routes using only the standard
library — no instrumented build, no profiler, one peer, ten seconds. It reads the thread's
CPU from `/proc`, which `top -H -p <pid>` shows equally well.

```
raptoreumd -regtest -datadir=/tmp/spin -listen=1 -port=19940
contrib/devtools/repro-socket-busyloop.py --port 19940 --pid $(pidof raptoreumd) --mode sendqueue
```

On an AMD Ryzen 9 3900X

| route | node options | before | after |
|---|---|---|---|
| send queue | **none** | **99.9%** of a core | 6.8% |
| pause | `-maxreceivebuffer=1` | 99.3% of a core | 5.1% |

`--mode sendqueue` is the RTM-specific route: it sends pings and never reads the replies, so
the node's send queue to it stays non-empty while its socket stays readable. It needs no
node options.

The repro is a review aid, not something that needs to live in the tree. The permanent
regression guard is the unit test in `src/test/net_tests.cpp`, which is mutation-verified and
runs in CI. If you would rather not carry a standalone DoS script under `contrib/`, drop it
before merge — the fix and the unit test stand without it.

## The fix

Two parts. It is wrong with only the first.

**1. Make the skip-wait test agree with the receive path.** `HasUnpausedReceivableNode()`
counts a receivable node as work only when the receive path would actually drain it. Both
sites carry a comment saying the two must stay in agreement, because that is the whole bug.

**2. Wake the socket thread when the pause clears.** The busy-loop was also how a paused peer
got re-examined. With it gone, nothing wakes the socket thread when `ProcessMessages` clears
`fPauseRecv`; the fallback wait is `SELECT_TIMEOUT_MILLISECONDS`, 500 ms with the wakeup pipe
enabled, and the sockets are edge-triggered, so data already buffered raises no fresh event
to cut that short.

The script also reports how much it managed to write; the socket is blocking, so that tracks
how fast the node drains it, and it shows why both parts are needed:

| | socket thread CPU | drained |
|---|---|---|
| before | full core | 20.9 MB/s |
| predicate only | near idle | **9.0 MB/s** |
| predicate + wakeup | near idle | 20.9 MB/s |

Shipping the predicate alone would have traded a visible problem for an invisible one.

`WakeSelect()` is called outside `cs_vProcessMsg`: writing the wakeup pipe under a lock the
socket thread contends on invites a deadlock.

## Relationship to the Dash fix

Dash fixed the pause route in `cb6afcbe5fea`, released in v23.1.7, describing the same
mechanism.

Their predicate tests only `fPauseRecv`, and their receive path does not require an empty
send queue. Ours does, so applying their version here leaves the send-queue route spinning.
That is why the predicate in this PR tests all three conditions rather than one.

The unit test encodes that. It covers each of the three conditions separately and checks that
one undrainable node does not mask a drainable one. It is mutation-verified: replacing the
send-queue term with `true`, which reduces the predicate to Dash's form, makes it fail.

## Why this is worth fixing before any throughput work

Two reasons that apply to any change raising sustained transaction volume, including the
transaction decoupling work being discussed.

**The trigger gets more common, not less.** The send-queue route fires when a node has bytes
queued to a peer whose socket also has data waiting. More transactions in flight means more
announcements queued per peer, more of the time. A condition that is already easy to reach
becomes close to permanent.

**It makes capacity measurements unreliable.** A node in this state stops making progress for
reasons that have nothing to do with its capacity, while still looking busy — full CPU, no
errors, no log output. On a core-constrained node the spinning socket thread contends with the message handler, so a
ceiling measured there can be a measurement of this bug rather than of capacity. On a host with
spare cores we measured no throughput difference: reverting this fix took the socket thread
from 12% to 84% of a core and left ingestion and relay unchanged within noise, because the spin
is on the socket thread while the bottleneck is the message handler. That is worth removing before
anyone sizes a design against numbers from this code.

## One thing to expect

This removes the busy-loop, not every way a connection can stop making progress. A peer can
still end up undrainable for longer than it should, and after this change that shows as a
quiet wait rather than a pinned core — so a stall seen under load after applying this is not
necessarily this bug returning. That path involves the send-readiness flag and edge-triggered
sockets, and is worth a separate look; it is not touched here.


JSanchezFDZ (approved):
Revisión de `469788e4f90b2afbc35ba88dbe6a47aadf8f81fb`: diff completo, vida de los nodos, escritores del estado de recepción/envío y backends de espera. **Aprobación del cambio C++: no encuentro una regresión bloqueante.** El predicado excluye nodos que no pueden avanzar y el despertar al retirar la pausa evita perder el progreso.

**Validación realizada**

- Volví a ejecutar `test_raptoreum --run_test=net_tests/socket_handler_skips_undrainable_nodes --log_level=test_suite` utilizando el artefacto Ubuntu 22 de [CI 35074172452](https://github.com/Raptor3um/raptoreum/actions/runs/35074172452), cuyo `head_sha` coincide exactamente: salida 0, `No errors detected`.
- Compilé las funciones reales extraídas `HasUnpausedReceivableNode` y `WakeSelect` en una prueba aislada: pasan las ocho combinaciones del predicado, mapas mixtos y el despertar anterior/durante `epoll_wait`. El resto del nodo se sustituye por una estructura mínima; no es un ensayo del daemon completo.
- Revisé cierre y referencias de nodos, orden de locks y vaciado de la cola de envío. No medí CPU/MB/s del daemon ni ejecuté macOS/FreeBSD.

**Dos correcciones no bloqueantes en el reproductor nuevo**

1. **P3 — Magic de mainnet/testnet incorrecto**, [repro-socket-busyloop.py:66](https://github.com/Raptor3um/raptoreum/blob/469788e4f90b2afbc35ba88dbe6a47aadf8f81fb/contrib/devtools/repro-socket-busyloop.py#L66). El script utiliza `bf0c6bbd`/`cee2caff`; este HEAD exige `72746d2e`/`7472746d`. El peer se desconecta por cabecera inválida y esos modos no generan la carga prevista. Contrasté el `frame()` real con los parámetros reales; regtest sí coincide. Corregir ambos valores o limitar explícitamente el script a regtest.
2. **P3 — Un fallo de envío puede clasificarse como éxito**, [líneas 140–141](https://github.com/Raptor3um/raptoreum/blob/469788e4f90b2afbc35ba88dbe6a47aadf8f81fb/contrib/devtools/repro-socket-busyloop.py#L140). El worker absorbe `OSError`; con CPU baja, el hilo principal puede imprimir `near 0 -> fixed` y devolver 0 aunque no haya enviado carga. Ejecuté el `main()` real con dobles de socket/CPU: cero bytes enviados y retorno 0. Propagar el error y marcar el ensayo como inválido, comprobando también el saludo.

**Cobertura y documentación**

El test unitario nuevo cubre el predicado, pero no detectaría la eliminación del despertar al despausar: añadir esa transición completaría la regresión. La afirmación de que todos los sockets usan notificación por flanco necesita matizarse: `kqueue` ya registra `EV_ADD` sin `EV_CLEAR` en la base. Es una limitación preexistente, no un fallo introducido aquí. Windows conserva su sondeo; mi comprobación de despertar cubre Linux/POSIX.

drmancini:
Thanks for the review. Addressed the two P3s and the two suggestions; also found and fixed a related issue on kqueue platforms while verifying. All pushed to this branch.

**1. Magic bytes (`repro-socket-busyloop.py`)** -- fixed: `main` is now `72746d2e`, `test` is `7472746d`, matching `CMainParams`/`CTestNetParams` in `chainparams.cpp`. `regtest` was already correct.

**2. Swallowed `OSError`** -- fixed: a worker socket failure before `stop` is requested now aborts the run with an explicit message instead of printing a misleading "near 0 -> fixed". Also added a handshake check and cleared a leftover 20s connect timeout that would otherwise misreport a legitimate long run past ~20s as "interrupted".

**3. "Sockets are edge-triggered" claim** -- reworded, both the code comment and the commit message. Confirmed against the actual `EV_SET` calls in `net.cpp`: kqueue's `EVFILT_READ` registration had no `EV_CLEAR`, so it was level-triggered there, not edge-triggered as originally claimed.

**4. Wakeup-transition test coverage** -- added `process_messages_wakes_select_on_unpause` in `denialofservice_tests.cpp`, wiring a private pipe into a real `CConnman`/`PeerLogicValidation` pair so the test observes `WakeSelect()` actually firing. Mutation-verified: deleting the `WakeSelect()` call makes it fail.

**5. kqueue busy-loop, found while checking (3) -- now fixed too:** the predicate/wakeup fix above doesn't close the loop on kqueue (BSD/macOS): `RegisterEvents` registered `EVFILT_READ` with `EV_ADD` only, no `EV_CLEAR`, unchanged since kqueue support was added in 2020 -- an asymmetry with the `EVFILT_WRITE` filter two lines below it, which already had `EV_CLEAR`. Added it. Verified on real macOS hardware (arm64): **99.5% CPU before, near idle after**, same `sendqueue`-mode workload as the reproducer script.

PR #482, opened 2026-09-17
[1/9] contrib: add raptoreum_hash, a GhostRider module for the test framework

drmancini:
The functional test framework builds and mines blocks in Python. Without a hash
function that matches consensus it cannot produce a header the node will accept, so
every test that mines a block by hand is unrunnable.

This adds `contrib/ghostrider-py`, a small CPython extension that binds Core's own
GhostRider implementation rather than reimplementing it. There is exactly one hash
function in the tree and the tests now use it.

No `src/` change. Nothing outside `contrib/` is touched.

The commit body explains the `PY_SSIZE_T_CLEAN` requirement: the `y#` argument format
needs that macro defined before `Python.h`, Python 3.10 through 3.12 raise
`SystemError` at call time without it, and 3.13 relaxed the rule. A module built and
smoke-tested only on 3.13 therefore looks correct while failing on every version the
rest of the suite can run on.

**Suite at this point:** still cannot run. The runner gates on a config key the build
does not emit; that is fixed in PR 4.

---

Part 1 of 9, replacing #470. First in the series. Targets `develop`; until the earlier parts merge, this diff also shows their commits, so review the commits rather than the combined diff.

---

Same change as #471, targeting `ft/breaking-up` instead of `develop`.


drmancini:
Closing as a duplicate. The same change is open as #471 against `develop`, review is happening there, and the series will reach `ft/breaking-up` from that side. The head branch is untouched, so nothing here is lost — happy to reopen against this base if it turns out to be more convenient.

PR #483, opened 2026-09-17
[2/9] test: remove tests that no longer describe this node

drmancini:
Three kinds of dead weight, one commit each:

- Seven stale copies of tests that were renamed upstream. Both names exist in the
  tree; the old copies were never updated and test nothing current.
- The governance test. Smartnodes do not run governance on Raptoreum; it is a Dash
  relic.
- The block reward reallocation test, which encodes Dash's reallocation schedule
  rather than this chain's.

Deletions only, apart from one line in `test_runner.py`.

**Suite at this point:** still cannot run, same gate as above.

---

Part 2 of 9, replacing #470. Builds on #471. Targets `develop`; until the earlier parts merge, this diff also shows their commits, so review the commits rather than the combined diff.

---

Same change as #472, targeting `ft/breaking-up` instead of `develop`.


drmancini:
Closing as a duplicate. The same change is open as #472 against `develop`, review is happening there, and the series will reach `ft/breaking-up` from that side. The head branch is untouched, so nothing here is lost — happy to reopen against this base if it turns out to be more convenient.

PR #484, opened 2026-09-17
[3/9] regtest: register test-only quorum types

drmancini:
Two problems stop a regtest chain from ever forming a quorum.

The test quorum types are not registered on regtest, so nothing can sign. And
`UpdateLLMQParams` rescales quorum parameters to the smartnode count on every block,
which overwrites whatever size a test chose through `-llmqtestparams` and re-registers
production-sized types that a handful of local nodes can never fill.

This registers the test types on regtest and makes `UpdateLLMQParams` return early
there. Scaling to smartnode count remains a mainnet and testnet concern, untouched.

This is the only consensus-adjacent code in the series, and it is inside a regtest
branch. Mainnet and testnet behaviour is unchanged.

**Suite at this point:** unit tests pass; the functional suite still cannot run.

---

Part 3 of 9, replacing #470. Builds on #472. Targets `develop`; until the earlier parts merge, this diff also shows their commits, so review the commits rather than the combined diff.

---

Same change as #473, targeting `ft/breaking-up` instead of `develop`.


drmancini:
Closing as a duplicate. The same change is open as #473 against `develop`, review is happening there, and the series will reach `ft/breaking-up` from that side. The head branch is untouched, so nothing here is lost — happy to reopen against this base if it turns out to be more convenient.

PR #485, opened 2026-09-17
[4/9] test: make the functional suite runnable

drmancini:
The commit that turns the suite on.

- `test_runner.py` gated the whole run on a configuration key the build never emits.
  With the key absent it exited zero having run nothing, which reads as success. This
  removes the gate.
- `authproxy.py` and `wallet_util.py` are taken from upstream rather than carrying
  local drift.
- The framework is adapted to Raptoreum: chain parameters, the extra RPC surface, and
  the block and address helpers.

**Suite at this point:** 34 passed, 86 failed, 2 skipped.

That failure count is the point of this PR, not a regression. Before it the suite
reported success while running nothing. The 86 failures are pre-existing and become
visible here; PR 5 repairs them.

---

Part 4 of 9, replacing #470. Builds on #473. Targets `develop`; until the earlier parts merge, this diff also shows their commits, so review the commits rather than the combined diff.

---

Same change as #474, targeting `ft/breaking-up` instead of `develop`.


drmancini:
Closing as a duplicate. The same change is open as #474 against `develop`, review is happening there, and the series will reach `ft/breaking-up` from that side. The head branch is untouched, so nothing here is lost — happy to reopen against this base if it turns out to be more convenient.

PR #486, opened 2026-09-17
[5/9] test: repair the suite

drmancini:
The bulk of the work, grouped by subject so each commit can be read on its own:
subsidy and founder fee assumptions, activation read from the update system rather
than BIP9, peer sync driven by the mocked clock, HD wallet assumptions, soft-fork
tests rewritten around rules that are always on, and then the rpc, wallet, feature,
mempool, p2p, tool, quorum, smartnode and evo tests in turn.

Five small `src/` fixes are included, each in its own commit alongside the test that
proves it:

| fix | file |
|---|---|
| separate the message caption from the message | `src/noui.cpp` |
| validate `-llmq-qvvec-sync` at startup | `src/init.cpp` |
| register the `-asmap` argument again | `src/init.cpp` |
| emit the `ischange` field `getaddressinfo` documents | `src/wallet/rpcwallet.cpp` |
| correct the "specified data directory" error message | `src/util/system.cpp` |

Two assertions in `interface_bitcoin_cli.py` are commented out rather than weakened,
each with a comment saying why. One is restored in PR 8 once the code it depends on is
fixed. The other concerns `-getinfo` fields this client does not emit and is a
permanent divergence.

**Suite at this point:** 111 passed, 8 failed, 3 skipped.

The eight are known and none is caused by this PR. `rpc_deriveaddresses` (twice, plain
and `--usecli`) and `rpc_named_arguments` need code fixes that land in PR 8.
`wallet_createwallet` (twice) is named by upstream's runner index but the file itself
has never existed in this fork; PR 7 adds it. `feature_block`,
`feature_llmq_connections` and `rpc_masternode` each appear at this stage only and at
no other, which is contention at twelve parallel jobs.

---

Part 5 of 9, replacing #470. Builds on #474. Targets `develop`; until the earlier parts merge, this diff also shows their commits, so review the commits rather than the combined diff.

---

Same change as #475, targeting `ft/breaking-up` instead of `develop`.


drmancini:
Closing as a duplicate. The same change is open as #475 against `develop`, review is happening there, and the series will reach `ft/breaking-up` from that side. The head branch is untouched, so nothing here is lost — happy to reopen against this base if it turns out to be more convenient.

PR #487, opened 2026-09-17
[6/9] rpc, llmq and wallet fixes found by the suite

drmancini:
Six defects, each with the test that catches it:

- `listsinceblock` did not filter conflicted transactions by depth.
- `smartnode payments` ignored its `count` argument.
- `logging` listed category aliases rather than the real categories.
- `getrpcinfo` did not return the `logpath` field it documents.
- Connections to quorums that fall outside the active window were never closed.
- The asset RPCs took `cs_main` before `cs_wallet`, the opposite order to the rest of
  the wallet, which is a lock inversion.

**Suite at this point:** 117 passed, 5 failed, 3 skipped: the same
`rpc_deriveaddresses`, `rpc_named_arguments` and `wallet_createwallet` entries, all
resolved by PR 7 and PR 8. The three single-stage flakes above do not recur.

---

Part 6 of 9, replacing #470. Builds on #475. Targets `develop`; until the earlier parts merge, this diff also shows their commits, so review the commits rather than the combined diff.

---

Same change as #476, targeting `ft/breaking-up` instead of `develop`.


drmancini:
Closing as a duplicate. The same change is open as #476 against `develop`, review is happening there, and the series will reach `ft/breaking-up` from that side. The head branch is untouched, so nothing here is lost — happy to reopen against this base if it turns out to be more convenient.

PR #488, opened 2026-09-17
[7/9] test: cover what nothing reached

drmancini:
New tests, no behaviour changes:

- Asset RPCs, what they accept and what they refuse.
- Future transactions.
- The founder fee.
- The query RPCs no test reached at all.
- Upstream tests the fork never took.
- `wallet_createwallet.py`, which was simply missing.

Three repairs to existing tests are included because they are coverage gaps rather
than failures: `feature_addressindex` picked an output at random, several tests
half-covered the RPC they named, and `create_block` now takes the node so hand-built
blocks are valid rather than accidentally so.

One of those repairs is worth calling out. `rpc_scantxoutset` pays its three scan
targets to the wallet's own addresses, so the thirteen sends that follow could select
and spend those very outputs, leaving the totals short by some subset of 0.001, 0.002
and 0.004. Measured at one failure in twenty-four runs. The outputs are now locked as
they are created, and their presence in the UTXO set is asserted before the wallet is
deleted, so a recurrence names its cause rather than surfacing as a wrong total forty
lines later. Ninety-six runs clean afterwards.

**Suite at this point:** 132 passed, 5 failed, 3 skipped. `rpc_deriveaddresses`
(twice) and `rpc_named_arguments` wait on PR 8. `wallet_address_queries` waits on the
`importelectrumwallet` fix in PR 8. `feature_llmq_is_cl_conflicts` appears at this
stage only.

---

Part 7 of 9, replacing #470. Builds on #476. Targets `develop`; until the earlier parts merge, this diff also shows their commits, so review the commits rather than the combined diff.

---

Same change as #477, targeting `ft/breaking-up` instead of `develop`.


drmancini:
Closing as a duplicate. The same change is open as #477 against `develop`, review is happening there, and the series will reach `ft/breaking-up` from that side. The head branch is untouched, so nothing here is lost — happy to reopen against this base if it turns out to be more convenient.

PR #489, opened 2026-09-17
[8/9] wallet and rpc fixes found by the new coverage

drmancini:
- `echo` could not echo. Its `RPCHelpMan` declares an empty argument list and the body
  validated against it, so every call with an argument threw the help text, while the
  dispatch table registers `arg0` through `arg9`. The registration and the help
  contradicted each other. This restores upstream's body. The assertion PR 5 commented
  out is restored in the same commit.
- `deriveaddresses` did not accept a `[begin,end]` range.
- `importprivkey` discarded an existing label.
- `-fallbackfee=0` paid the floor instead of being honoured.
- `importelectrumwallet` read the extension with `find_last_of(".")` over the whole
  argument, so a path under a directory called `2.0.3` or `.raptoreum` supplied one and
  a file with no extension at all was reported as having the wrong one. It now asks
  `fs::path` for the extension, as the same function already does forty lines further
  down. This is the only site in the tree that reads an extension this way. The test
  creates the dotted directory itself rather than relying on the build path, so the
  case fails on the old code everywhere rather than only where the tree happens to sit
  under a dot.

**Suite at this point:** 139 passed, 0 failed, 3 skipped. With `pyzmq` installed,
142 passed, 0 failed, nothing skipped.

---

Part 8 of 9, replacing #470. Builds on #477. Targets `develop`; until the earlier parts merge, this diff also shows their commits, so review the commits rather than the combined diff.

---

Same change as #478, targeting `ft/breaking-up` instead of `develop`.


drmancini:
Closing as a duplicate. The same change is open as #478 against `develop`, review is happening there, and the series will reach `ft/breaking-up` from that side. The head branch is untouched, so nothing here is lost — happy to reopen against this base if it turns out to be more convenient.

PR #490, opened 2026-09-17
[9/9] test: run by default what upstream runs by default

drmancini:
Aligns `test_runner.py`'s default set with upstream's, and removes the one genuine
duplicate entry (`p2p_unrequested_blocks.py`). Entries that look duplicated are the
same script under different flags, `--usecli`, `--ipv4` and `--spork21`, and are kept.

**Suite at this point:** 146 passed, 0 failed, 3 skipped.

With `pyzmq` installed the suite is 149 tests with nothing skipped. A clean run is
147 passed and 2 failed, both of them intermittent and neither introduced by this
series. They are described below.

---

Part 9 of 9, replacing #470. Builds on #478. Targets `develop`; until the earlier parts merge, this diff also shows their commits, so review the commits rather than the combined diff.

---

Same change as #479, targeting `ft/breaking-up` instead of `develop`.


drmancini:
Closing as a duplicate. The same change is open as #479 against `develop`, review is happening there, and the series will reach `ft/breaking-up` from that side. The head branch is untouched, so nothing here is lost — happy to reopen against this base if it turns out to be more convenient.

PR #492, opened 2026-09-17
validation: don't copy the asset cache for every tx

drmancini:
AcceptToMemoryPoolWorker copied the entire global CAssetsCache (CAssetsCache assetsCache = *passetsCache.get()) on every transaction, before CheckSpecialTx. That copy deep-copies three std::maps that grow with the number of confirmed assets, yet CheckSpecialTx consults it only for the three asset transaction types (NEW_ASSET / UPDATE_ASSET / MINT_ASSET) and returns at its first line for every other transaction without dereferencing it. Every ordinary payment therefore paid an O(confirmed-assets) copy it never used.

The in-memory cache is capped at MAX_CACHE_ASSETS_SIZE (2500) by LoadAssets at startup and grows from there over uptime. On a mainnet node (3,439 confirmed assets, ~2,500 loaded) the copy costs ~2-3 ms per transaction, which alone caps single-thread mempool acceptance at a few hundred tx/s regardless of transaction type.

Construct the working copy only for the three asset transaction types and pass nullptr otherwise; CheckSpecialTx dereferences the pointer only for those types, so the change is behaviour-preserving. Payment-transaction acceptance becomes independent of the confirmed-asset count and of node uptime; asset transactions are unaffected.

bench/assets_cache_copy.cpp measures the copy cost as a function of asset count. A regtest before/after at 2,500 in-memory assets: the copy path on a normal payment drops from ~1.16 ms to ~0.02 ms. See docs/asset-cache-drag.md.

---

Same change as #481, targeting `ft/breaking-up` instead of `develop`.


drmancini:
Closing as a duplicate. The same change is open as #481 against `develop`, review is happening there, and the series will reach `ft/breaking-up` from that side. The head branch is untouched, so nothing here is lost — happy to reopen against this base if it turns out to be more convenient.

